PR #29, opened 2021-12-21
Update torchrec's submodule fbgemm commit id

s4ayub:
Summary:
**context**
fbgemm was failing to install out of torchrec's setup.py command because the THC.h header file was being included in the fbgemm repo but it was removed.

This was fixed here: D32872476 but was not reflected in torchrec's checkout of fbgemm as that has to be manually updated.

**this diff:**
This updates the commit id to be the latest.

I'll take a look at how to automate this process if possible. The installation process can break if our version of fbgemm gets too stale.

Reviewed By: jianyuh

Differential Revision: D33251641


s4ayub:
merged with shipit

PR #53, opened 2022-02-08
Create docs.yml

brianjo:
This action should build the docs and deploy them to the gh-pages branch. We can serve these through GitHub pages from settings.


joshuadeng (changes requested):
Hey Brian, we currently are using the internal repo as source of truth. If you can submit a diff internally I can approve it

PR #136, opened 2022-03-08
Fix dlrm feature-interaction triu_indices

samiwilf:
Change triu_indices to match  FB dlrm oss (https://github.com/facebookresearch/dlrm/blob/9c2fda79afbc09e277c17e420bffe510125b4f70/dlrm_s_pytorch.py#L491-L493) 
and Nvidia dlrm (https://github.com/NVIDIA/DeepLearningExamples/blob/690233ca8eaf11b9fb7b9dfdb5346ce149f17cea/PyTorch/Recommendation/DLRM/dlrm/nn/interactions.py#L50-L53)

### DIFF
--- a/torchrec/models/dlrm.py
+++ b/torchrec/models/dlrm.py
@@ -187,8 +187,8 @@ class InteractionArch(nn.Module):
     def __init__(self, num_sparse_features: int) -> None:
         super().__init__()
         self.F: int = num_sparse_features
-        self.triu_indices: torch.Tensor = torch.triu_indices(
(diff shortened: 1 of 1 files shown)


s4ayub:
@samiwilf  did this change have any impact on auroc?

samiwilf:
@s4ayub AUROC went up, but it still doesn't reach 80 within 1 epoch of training.  I'm continuing looking into why.

samiwilf:
The triu indices are actually right, just different from other implementations.

PR #149, opened 2022-03-12
Fix test failed by multiple process group initialization

KanchiShimono:
Summary:
Unit tests are currently failed, because process group is initialized multiple times without cleanup.
I also updated the test case to work both nccl and gloo backend.

### DIFF
--- a/torchrec/distributed/tests/test_model_parallel.py
+++ b/torchrec/distributed/tests/test_model_parallel.py
@@ -344,7 +344,7 @@ class ModelParallelTest(ModelParallelTestShared):
         torch.cuda.device_count() <= 1,
         "Not enough GPUs, this test requires at least two GPUs",
     )
-    def test_sharding_nccl_module_as_top_level(self) -> None:
+    def test_top_lovel_shardable_module_is_sharded_nccl(self) -> None:
         backend = "nccl"
         device = torch.device("cuda:0")
         pg = init_distributed_single_host(
@@ -379,7 +379,7 @@ class ModelParallelTest(ModelParallelTestShared):
             dist.destroy_process_group(_CROSS_PG)
         dist.destroy_process_group(pg)
 
-    def test_sharding_gloo_module_as_top_level(self) -> None:
+    def test_top_lovel_shardable_module_is_sharded_gloo(self) -> None:
         backend = "gloo"
         device = torch.device("cpu")
         pg = init_distributed_single_host(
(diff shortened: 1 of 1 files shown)


torchrec/distributed/tests/test_model_parallel.py
```diff
@@ -335,22 +340,54 @@ def test_sharding_gloo_dp(
...
-        os.environ["LOCAL_WORLD_SIZE"] = "1"
-        os.environ["MASTER_ADDR"] = str("localhost")
-        os.environ["MASTER_PORT"] = str(get_free_port())
-        os.environ["NCCL_SOCKET_IFNAME"] = "lo"
+    @unittest.skipIf(
+        torch.cuda.device_count() <= 1,
+        "Not enough GPUs, this test requires at least two GPUs",
+    )
+    def test_top_lovel_shardable_module_is_sharded_nccl(self) -> None:
+        backend = "nccl"
```
colin2328, line 348: most of our tests run NCCL for GPU, and GLOO for CPU. This is done automatically.

We are currently only running tests on GPU for github CI, but we hope to fix this in the future (e.g adding a github action to run unit tests on CPU as well every PR/merge). (internally to Meta AIs servers, we run both CPU and GPU testss)

 If you are interested in contributing to the repo, you're welcome to submit a PR which adds this github action to run on CPU!

colin2328:
@KanchiShimono thanks for jumping in to fix this!

We actually have an internal  diff which is being exported now to fix this  issue (you're right in the root cause ) - see https://github.com/pytorch/torchrec/pull/153

torchrec/distributed/tests/test_model_parallel.py
```diff
@@ -335,22 +340,54 @@ def test_sharding_gloo_dp(
...
+                    pooling=PoolingType.SUM,
+                ),
+            ],
+        )
+
+        model = DistributedModelParallel(ebc, device=device)
+
+        self.assertTrue(isinstance(model.module, ShardedEmbeddingBagCollection))
+
+        if _INTRA_PG is not None:
```
colin2328, line 376: btw, these PGs won't be created in this test (only for hierarchical sharding style )

torchrec/distributed/tests/test_model_parallel.py
KanchiShimono: Sounds good!
I'll try adding tests on CPU.

torchrec/distributed/tests/test_model_parallel.py
KanchiShimono: Thanks for teaching me!
I gained a deeper understanding.

KanchiShimono:
@colin2328 
Thanks for reply and review!

TorchRec is most interesting project for me!
I'm very excited about your teams shared such amazing library as OSS!
So I wanna contribute this project continuously.

The issue is fixed by #153, so I close this PR.

PR #171, opened 2022-03-24
GHA workflow for CPU unit testing

KanchiShimono:
Summary:
Adding GitHub Actions workflow for CPU unit testing triggered by each PR/merge.

### DIFF
--- a/.github/workflows/unittest_ci_cpu.yml
+++ b/.github/workflows/unittest_ci_cpu.yml
@@ -7,94 +7,12 @@ on: [push, pull_request]
 
 jobs:
 
-  # build on cpu hosts and upload to GHA
-  build_on_cpu:
+  # build and test on cpu hosts
+  build_test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [linux.2xlarge]
-    steps:
-    # Checkout the repository to the GitHub Actions runner
-    - name: Check ldd --version
-      run: ldd --version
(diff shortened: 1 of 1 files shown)


.github/workflows/unittest_ci_cpu.yml
```diff
@@ -0,0 +1,206 @@
...
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: CPU Unit Test CI
+
+on: [push, pull_request]
+
+jobs:
+
+  # build on cpu hosts and upload to GHA
+  build_on_cpu:
```
colin2328: when running unit tests on GPUs, we separate the GHA steps into two phases:
1) build on CPU
2) test on GPU

since we are running it all on CPU, we don't need to do this. See https://github.com/pytorch/torchrec/blob/main/.github/workflows/release_build_cpu.yml#L12 for an example

colin2328:
thanks @KanchiShimono  for starting this!

colin2328 (changes requested):
@KanchiShimono putting it back to you, please combine the two phases since we dont need to separate

.github/workflows/unittest_ci_cpu.yml
KanchiShimono: @colin2328 
Thank you for your review!
I fixed to combine build and test phases.

PR #180, opened 2022-03-26
Reinstate ufmt pre commit hook

zhiqwang:
Since https://github.com/omnilib/ufmt/issues/56 is resolved, so I re-enable this hook, and upgrade it to v1.3.2 as https://github.com/pytorch/vision/pull/5460 .


colin2328:
@zhiqwang thanks for your contribution.

Our CI process involves pulling into internal for now, so I'll do that now

PR #214, opened 2022-04-11
Fix topology not properly set when running dlrm_main multinode

Luo-Liang:


colin2328:
thanks @Luo-Liang -> I think this isn't relevant anymore, sorry formissing the PR

PR #295, opened 2022-05-05
add external fbgemm lib path 

pengwa:
If skip_fbgemm is enabled,  the fbgemm submodule in third_party will not be built and installed as dependency. Instead, some pre-installed fbgemm lib is used. So I think we should allow user to pass the external fbgemm lib path before doing the build.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,9 +126,7 @@ def main(argv: List[str]) -> None:
             print(out)
 
         # the path to find all the packages
-        fbgemm_install_base = glob.glob(
-            fbgemm_gpu_install_dir
-        )[0]
+        fbgemm_install_base = glob.glob(fbgemm_gpu_install_dir)[0]
         packages.extend(find_packages(fbgemm_install_base))
         # to include the fbgemm_gpu.so
         fbgemm_gpu_package_dir = glob.glob(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -60,6 +60,12 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
         action="store_true",
         help="if fbgemm_gpu will be installed with cpu_only flag",
     )
+    parser.add_argument(
+        "--fbgemm_gpu_dir",
+        type=str,
+        default="third_party/fbgemm/fbgemm_gpu",
+        help="the directory of external fbgemm_gpu path. Only applicable when skip_fbgemm is enabled.",
+    )
```
zyan0: I would suggest default to None here

setup.py
```diff
@@ -89,6 +95,7 @@ def main(argv: List[str]) -> None:
 
     packages = find_packages(exclude=("*tests",))
     fbgemm_gpu_package_dir = []
+    fbgemm_install_dir = os.path.join(args.fbgemm_gpu_dir, "_skbuild/*/cmake-install")
```
zyan0: If fbgemm_gpu_dir_override is set, then use that, otherwise use the default one

zyan0:
thanks! lgtm! could you fix the pre-commit failure? (about Python format)

https://github.com/pytorch/torchrec/runs/6331006917?check_suite_focus=true

PR #402, opened 2022-06-07
Criteo preprocessing improvements

janekl:
I'm using the fact how Criteo dataset looks like to improve preprocessing:
* there is no need to cast sparse features to `str` first  as `csv.reader` [here](https://github.com/pytorch/torchrec/blob/50ff0973d5d01ec80fe36ba5f1d524c92c799836/torchrec/datasets/utils.py#L216) produces a list strings 
* we can replace missing values -- empty strings -- to zeros right away and then cast features to int avoiding try/catch block in `safe_cast` helper function.

This results in a decent 15% speedup according to my measurements. It is a non-negligible time saving as the preprocessing is a long job. Moreover, I think that this is slightly more transparent in how .npy files are produced.

### DIFF
--- a/torchrec/datasets/criteo.py
+++ b/torchrec/datasets/criteo.py
@@ -199,10 +199,11 @@ class BinaryCriteoUtils:
         """
 
         def row_mapper(row: List[str]) -> Tuple[List[int], List[int], int]:
-            label = safe_cast(row[0], int, 0)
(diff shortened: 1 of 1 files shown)


YLGH:
Thanks @janekl, changes LGTM, shipping by EOD.

BTW, we're experimenting with using nvtabular/other methods of preprocessing, that can hopefully drive the long runtime down.

PR #404, opened 2022-06-07
Change module walk to use `_fsdp_wrapped_module` instead of `module`

awgu:
Summary: This fixes the TorchRec module walk after https://github.com/pytorch/pytorch/pull/78671 changed `FullyShardeDataParallel.module` to return the inner wrapped module instead of the `FlattenParamsWrapper`.

Differential Revision: D36941682


awgu:
Change was merged without this PR.

PR #427, opened 2022-06-13
add some missing inference mode guards

zyan0:
Summary: make sure each thread touching tensor has a inference guard

Differential Revision: D37100943


houseroad (approved):
Looks good

PR #448, opened 2022-06-17
Fix Nightly CPU installation in README

KanchiShimono:
As title.
Fix package name and link of Nightly CPU binary.


YLGH:
good catch, thanks @KanchiShimono !

PR #478, opened 2022-06-27
Remove submodule

YLGH:
Since we no longer depend on FBGEMM/submodules, remove this


YLGH:
See https://github.com/pytorch/torchrec/actions/runs/2571363235

PR #481, opened 2022-06-28
[RFC] Dynamic Embedding with PS

zhuzilin:
Hi TorchRec team, here are some of our ideas on dynamic embeddings. Looking forward to discussing this with you :)


rfc/RFC-0001-dynamic-embedding-with-PS.md
```diff
@@ -0,0 +1,95 @@
...
+
+    These 2 approaches are a trade-off between network communication and computation. It is worth mentioning that the latter approach is more complex. And the second all-to-all operation in the first approach could be concurrent with the fetching from PS to GPU cluster.
+
+3. Will there be intrusive modifications to TorchRec?
+
+    To implement prefetching, i.e. preprocessing future data while training on the current ones, we need a synchronization mechanism. This is because we need to know which data are ready to train and which have finished training (ready to evict). Also, we could postpone the synchronization of fetching to just before calling `forward` on embedding modules. We may register hooks to the DMP model.
```
colin2328, line 95: at what point would this sync happen?

rfc/RFC-0001-dynamic-embedding-with-PS.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Summary
+In a social media scenario, it is common that the feeds to recommend and the number of users keep growing. As a result, users of TorchRec would expect their recommender system support training without estimating the upper bound of the number of feeds in advance. We would call the feature to support growing embedding size as dynamic embedding.
+
+We propose to solve dynamic embedding by attaching an external parameter server (PS) to TorchRec. TorchRec would automatically push/pull the features to/from PS during training. We also propose certain APIs for dynamic embedding, so that TorchRec could pipeline the network communication with the model computation and achieve maximum performance. 
+
+## Background
+With the rapid development of social media, recommender systems face a new challenge – the number of feeds to recommend grows daily and it is hard to train models with a fixed size embedding. Therefore, we propose a new feature to help companies train large recommendation models with dynamic embedding support and great performance.
```
colin2328, line 13: few qs on the background:
Overall, this would still be SPMD fully synchronous training, with the exception that the PS fetches (i.e cache miss) will be async?
What is the mechanism responsible for allocating memory? (today, the planner produces a shardingPlan which includes which params live where, including UVM, HBM, etc). Would the planner also be responsible for allocating params across HBM (for cache) and parameter servers?
Would sharding still happen with torchrec.distributed.DMP?

rfc/RFC-0001-dynamic-embedding-with-PS.md
```diff
@@ -0,0 +1,95 @@
...
+3. Online training matters. Users may hope to push the model trained with data of hotspots to the inference side within minutes so that their users could enjoy the most up-to-date content.
+
+From the above characteristics, it is hard to support dynamic embedding by starting training with a larger GPU cluster. The problem with this solution is that it is hard for users to estimate the size of the cluster to prevent a waste of resources and crash at the same time. Therefore, we propose to introduce an interface in TorchRec to allow attaching external storage components. Specifically, we suggest adding support for attaching PS to TorchRec cluster for several reasons:
+
+1. The remote PS and the embedding of TorchRec clusters form a hot-cold separation of the features naturally.
+2. There are mature industrial solutions for CPU PS that could support nearly unlimited storage and quickly send the training model weight to inference.
+3. PS could decouple the training and inference process, making it possible to migrate to TorchRec incrementally.
+4. The cost of CPU PS is normally much lower than GPU servers.
```
colin2328, line 45: Would trainers always be on GPU?

rfc/RFC-0001-dynamic-embedding-with-PS.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Background
+With the rapid development of social media, recommender systems face a new challenge – the number of feeds to recommend grows daily and it is hard to train models with a fixed size embedding. Therefore, we propose a new feature to help companies train large recommendation models with dynamic embedding support and great performance.
+
+## APIs
+
+We propose a wrapper API that could allow us to change the model and data loader together:
+
+```python
+with torchrec.contrib.DynamicEmbedding(
```
colin2328, line 20: why a context manager?

rfc/RFC-0001-dynamic-embedding-with-PS.md
```diff
@@ -0,0 +1,95 @@
...
+To implement this architecture, we need to clarify some implementation details:
+
+1. How to move the features fetched from PS into GPU embedding?
+
+    Because TorchRec supports multiple sharding schema, different segments of an embedding feature may be stored on different devices, when given a `cache_id`, we need to cut it into segments and move those segments to their actual locations in the ShardedTensor.
+    We may store features in PS according to the sharding schema, i.e. save segments of the feature separately, so that each device can only fetch the segment that will be in its shard.
+
+2. How to have a consistent cache index across devices?
+
+    During training, the same `cache_id` on different devices must be storing the feature with the same `global_id`. To ensure strong consistency, we could use let each device only maintain the cache info of the features that are stored locally, and communicate with other devices if cache info of other features is needed.
```
colin2328, line 77: I'm assuming this means locks/mutexes

rfc/RFC-0001-dynamic-embedding-with-PS.md
```diff
@@ -0,0 +1,95 @@
...
+
+    <img src="./RFC-0001-assets/cache-consistency.png" style="width:400px;" />
+
+    In the illustration above, each GPU gets a different data sample (data-parallel training), and each data sample has some ids that should be stored on GPU 0, 1, and 2 (represented by blue, orange, and green). We could first use an all-to-all operation to gather all ids with the same color to its corresponding GPU so that each GPU could deal with their ids. And after updating the cache, we could use another all-to-all operation to scatter the ids back (not drawn in the above illustration).
```
colin2328, line 81: will this all occur in the 'input_dist/output_dist' phase of the ShardedModule?

colin2328 (approved):
Thanks @zhuzilin . Approving this RFC and will land into the library. I have some questions, if you could address at some point.

My sense is that this is going to be a big change to the library. It may take us on our side some time to think through what we recommend regarding API, design decisions etc.

In the meantime, we propose that you make PRs to https://github.com/pytorch/torchrec/tree/main/torchrec/experimental/dynamic 
this will hopefully let you develop more quickly, but we may ask for changes.
Does this work for you?

rfc/RFC-0001-dynamic-embedding-with-PS.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. The remote PS and the embedding of TorchRec clusters form a hot-cold separation of the features naturally.
+2. There are mature industrial solutions for CPU PS that could support nearly unlimited storage and quickly send the training model weight to inference.
+3. PS could decouple the training and inference process, making it possible to migrate to TorchRec incrementally.
+4. The cost of CPU PS is normally much lower than GPU servers.
+
+**NOTE**: The PS here is only responsible for storing the embedding features, similar to KV storage like redis. All the model training computations (forward, backward, and optimizer) are done within the TorchRec GPU cluster.
+
+Under such a setting, the distributed embedding in GPU clusters would serve as a cache of the whole system. For each feature, it will have a `global_id`, representing its id in PS, and if it is currently stored in GPU embedding, a `cache_id`, representing its id in the GPU embedding. When a new data sample arrives, it needs to convert the `global_id` to `cache_id` and update the “cache” if necessary. The following is pseudocode of the preprocessing method:
```
colin2328, line 49: we will likely want to modify fbgemm_gpu nn.module / kernels for placement

rfc/RFC-0001-dynamic-embedding-with-PS.md
reyoung: Basically like the graph below,

image

There could be a background thread to preprocess data batch and fetch/evict correct parameters from/to parameter server. Since they are IO-intensive operators, they can be just asynchronous(e.g., python asyncio/C++ asio/CUDA memcpy async). In the background thread, we just launch asynchronous tasks and do not wait for them to complete until the data batch needs to be trained.

In another word, we just prefetch some data batches to pipeline `dynamic embedding` related I/O and normal training.

rfc/RFC-0001-dynamic-embedding-with-PS.md
reyoung: > Overall, this would still be SPMD fully synchronous training, with the exception that the PS fetches (i.e cache miss) will be async?

Yes.

In addition, saving embedding to ParamServer could be asynchronous, too. So, not only PS fetches, but also PS pushes(i.e., cache evict).

> What is the mechanism responsible for allocating memory? (today, the planner produces a shardingPlan which includes which params live where, including UVM, HBM, etc). Would the planner also be responsible for allocating params across HBM (for cache) and parameter servers?

1. Would the planner also be responsible for allocating params across HBM (for cache)?
  * No. 
  * For example
      * if we want a Embedding(name="feed_ids") be infinity size/`dynamic embedding`. We just config this embedding as `Embedding(name="feed_ids", num_embeddings=65536)`. The `num_embeddings` in this case, is just the `num_embeddings` in `TorchRec Models/HBM/UVM, etc`, and we reuse TorchRec's planner to sharding it into cluster.
      * And with a decorator `DynamicEmbedding(model, remote="redis://...", embedding_names=["feed_ids"])`, we just tell DynamicEmbedding extension to swap in/out `feed_ids` Embedding.

2. allocating params in parameter servers.
  * In ParameterServer, we just implement a method like `Fetch(item_id ) -> Optional<torch::Tensor>`
  * If the fetch returns empty, then we just initialize this cache row in `TorchRec Model`.
     * something like,  `model.state_dict()["xxx.xxx.feed_ids"].local_shards().tensor[row].normal_()` or `optimizer.state_dict()[...]..`

> Would sharding still happen with torchrec.distributed.DMP?

Yes. We do not change any sharding or distributed paradigm in `TorchRec`.

rfc/RFC-0001-dynamic-embedding-with-PS.md
reyoung: Not sure. We actually have not tested TorchRec on CPUs.  But it seems that the `DynamicEmbedding` RFC can fit training on CPUs also.

@zhuzilin maybe we should remove this line?

rfc/RFC-0001-dynamic-embedding-with-PS.md
reyoung: Since there are background threads, parameter server connection pools, and cache tables to manage.

rfc/RFC-0001-dynamic-embedding-with-PS.md
reyoung: In general, yes. The mapping between `cache_id` and `global_id` should be locked.

I think it could be a lock-free implementation, since generally it just an `std::vector<atomic<long>> cache_ids_to_global_ids`. 

Instead of synchronizing the map, we think to synchronize the async job is simpler. Each data batch feed to `TorchRec` just like 

```
struct DataSampleToTorchRec {
  std::future<void()> fetch_job_complete_;
  DataBacth transformed_data_batch_;
};
```

PR #503, opened 2022-07-06
Fix EmbeddingEnumerator error raising

zhuzilin:
The old exception mechanism will raise error when Enumerator cannot find a compute kernel for given sharding type, which will disallow user to set only compute kernel constraints. For instance, user may hope to set the compute kernel as UVM, which leads to:

```python
constraints = {
    "table": ParameterConstraints(
        compute_kernels=[EmbeddingComputeKernel.FUSED_UVM_CACHING.value]
    )
}
```
However, the `data_parallel` does not allow UVM, which would stop the training. This PR is trying to fix this issue.

Thank you for your time on this PR :)

### DIFF
--- a/torchrec/distributed/planner/enumerators.py
+++ b/torchrec/distributed/planner/enumerators.py
@@ -5,6 +5,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import logging
 import math
 from typing import Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


torchrec/distributed/planner/enumerators.py
```diff
@@ -197,10 +202,6 @@ def _filter_compute_kernels(
             ):  # always false for data_parallel
                 filtered_compute_kernels.remove(EmbeddingComputeKernel.DENSE.value)
 
-        if not filtered_compute_kernels:
-            raise RuntimeError(
```
joshuadeng, line 201: Instead of deleting this completely, let's make this `logger.warn`, and do the same in `_filter_sharding_types`, we don't want to raise 2 separate errors.

You can instantiate the logger with:
`import logging`
`logger: logging.Logger = logging.getLogger(__name__)`

joshuadeng (changes requested):
Thanks for the catch! Left some small follow up and it should be good to go

torchrec/distributed/planner/enumerators.py
zhuzilin: fixed.

joshuadeng (approved):
Awesome, thanks!

PR #504, opened 2022-07-06
Add fused_params arg to DMP and move fused_params to Sharder.shard()

zhuzilin:
This PR adds `fused_params` to args of DMP and move `fused_params` from the constructor of sharders to `Sharder.shard()`. There are several reasons for this change:

- `fused_params` contains important information of training, for examples, the optimizer type and its hyperparameters, so it is often a bug in the code when `fused_params` is not provided. However, for a user who is trying with the default sharders, it's easy to ignore the fact that DMP will create empty fused params in:
  ```python
  def get_default_sharders() -> List[ModuleSharder[nn.Module]]:
      return [
          cast(ModuleSharder[nn.Module], EmbeddingBagCollectionSharder()),
          cast(ModuleSharder[nn.Module], FusedEmbeddingBagCollectionSharder()),
          ...
      ]
  ```
  Therefore, this PR raises a warning when `fused_params` is not passed to DMP and change the code into:
  ```python
  class DMP(...):
    def __init__(self, ..., fused_params=None, ... sharders=None, ...):
        ...
        if fused_params is None:
            logging.warning("...!")
        if sharders is None:
            self.sharders = get_default_sharders()
        ...
        sharded_module = self.sharders[i].shard(module, fused_params, ...)
        ...
  ```
- It's unintuitive that when creating sharder, one should pass the optimizer info for all the embeddings it would shard. Moving `fused_params` to `shard()` would allow user to pass different parameter, e.g. `cache_load_factor`, to different sharded embeddings.

Thank you for your time on this PR :)

### DIFF
--- a/examples/bert4rec/bert4rec_main.py
+++ b/examples/bert4rec/bert4rec_main.py
@@ -494,9 +494,8 @@ def main(argv: List[str]) -> None:
         model = DMP(
             module=model_bert4rec,
             device=device,
-            sharders=[
-                cast(ModuleSharder[nn.Module], EmbeddingCollectionSharder(fused_params))
-            ],
+            sharders=[cast(ModuleSharder[nn.Module], EmbeddingCollectionSharder())],
+            fused_params=fused_params,
         )
         dense_optimizer = KeyedOptimizerWrapper(
             dict(model.named_parameters()),
@@ -518,7 +517,7 @@ def main(argv: List[str]) -> None:
                 "item_embedding"
             ] = torchrec.distributed.planner.ParameterConstraints(sharding_types=sharding_types)
             sharders = [
-                cast(ModuleSharder[nn.Module], EmbeddingCollectionSharder(fused_params))
+                cast(ModuleSharder[nn.Module], EmbeddingCollectionSharder())
             ]
             pg = dist.GroupMember.WORLD
             model = DMP(
@@ -529,7 +528,8 @@ def main(argv: List[str]) -> None:
                     world_size=world_size,
                     compute_device=device.type,
                 ),
-                constraints=constraints
+                constraints=constraints,
+                fused_params=fused_params,
             ).collective_plan(model_bert4rec, sharders, pg),
                 sharders=sharders,
             )
--- a/examples/nvt_dataloader/train_torchrec.py
+++ b/examples/nvt_dataloader/train_torchrec.py
@@ -231,7 +231,7 @@ def main(argv: List[str]):
     sharders = cast(
         List[ModuleSharder[nn.Module]],
         [
-            EmbeddingBagCollectionSharder(fused_params=fused_params),
+            EmbeddingBagCollectionSharder(),
         ],
     )
 
@@ -256,6 +256,7 @@ def main(argv: List[str]):
             ),
         ).collective_plan(train_model, sharders, pg),
         sharders=sharders,
+        fused_params=fused_params,
     )
 
     non_fused_optimizer = KeyedOptimizerWrapper(
--- a/examples/ray/train_torchrec.py
+++ b/examples/ray/train_torchrec.py
@@ -99,13 +99,14 @@ def train(
         "optimizer": OptimType.EXACT_ROWWISE_ADAGRAD,
     }
     sharders = [
-        EmbeddingBagCollectionSharder(fused_params=fused_params),
+        EmbeddingBagCollectionSharder(),
     ]
     # Distribute model across devices
     model = DistributedModelParallel(
         module=train_model,
         device=device,
         sharders=cast(List[ModuleSharder[nn.Module]], sharders),
+        fused_params=fused_params,
     )
 
     # Overlap comm/compute/device transfer during training through train_pipeline
--- a/examples/retrieval/two_tower_train.py
+++ b/examples/retrieval/two_tower_train.py
@@ -145,7 +145,7 @@ def train(
     }
     sharders = cast(
         List[ModuleSharder[nn.Module]],
-        [EmbeddingBagCollectionSharder(fused_params=fused_params)],
+        [EmbeddingBagCollectionSharder()],
     )
 
     # TODO: move pg to the EmbeddingShardingPlanner (out of collective_plan) and make optional
@@ -164,6 +164,7 @@ def train(
         module=two_tower_train_task,
         device=device,
         plan=plan,
+        fused_params=fused_params,
     )
 
     optimizer = KeyedOptimizerWrapper(
--- a/examples/torcharrow/run.py
+++ b/examples/torcharrow/run.py
@@ -89,8 +89,9 @@ def main(
         module=model,
         device=device,
         sharders=[
-            EmbeddingBagCollectionSharder(fused_params=fused_params),
+            EmbeddingBagCollectionSharder(),
         ],
+        fused_params=fused_params,
     )
 
     optimizer = KeyedOptimizerWrapper(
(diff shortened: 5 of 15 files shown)


colin2328:
@zhuzilin , thanks for the PR and for the description of the issue you're trying to solve for!

We agree, and we have some internal work (and planned work) to address these issues. 
Firstly, we are working on better decoupling of fusion/batching (i.e compute_kernel in torchrec) and sharding
As part of this, we're also making it easier to configure things like optimizer fusion (and optimizer hyperparams) in a more intuitive, pytorch-native way. Like you pointed out, the user can easily forget to pass in optimizer hyperparams and unintentially get different optimizer settings, resulting in bugs.

Our proposed solution is to allow the users to better specify the BatchedFused modules (including fused_params) first, and then pass these into DMP (to shard). https://github.com/pytorch/torchrec/issues/505 (just added) includes more information here.

Parts of this decomposition are already completed. See https://pytorch.org/blog/pytorch-1.12-new-library-releases/#prototype-single-process-batched--fused-embeddings release notes. You can see this new API (using the fuse_embedding_optimizer wrapper) here: https://github.com/pytorch/torchrec/blob/main/Torchrec_Introduction.ipynb. Can you try this out and see if it meets your needs?

In our opinion, a full decoupling is the way to go, vs passing in fused_params to DMP. (We're also rethinking what DMP could look like in the future). I'm going to reject PR for now, but if you think that what we have planned doesn't meet your needs, please comment. cc @YLGH  @divchenko

YLGH:
> @zhuzilin , thanks for the PR and for the description of the issue you're trying to solve for!
> 
> We agree, and we have some internal work (and planned work) to address these issues. Firstly, we are working on better decoupling of fusion/batching (i.e compute_kernel in torchrec) and sharding As part of this, we're also making it easier to configure things like optimizer fusion (and optimizer hyperparams) in a more intuitive, pytorch-native way. Like you pointed out, the user can easily forget to pass in optimizer hyperparams and unintentially get different optimizer settings, resulting in bugs.
> 
> Our proposed solution is to allow the users to better specify the BatchedFused modules (including fused_params) first, and then pass these into DMP (to shard). #505 (just added) includes more information here.
> 
> Parts of this decomposition are already completed. See https://pytorch.org/blog/pytorch-1.12-new-library-releases/#prototype-single-process-batched--fused-embeddings release notes. You can see this new API (using the fuse_embedding_optimizer wrapper) here: https://github.com/pytorch/torchrec/blob/main/Torchrec_Introduction.ipynb. Can you try this out and see if it meets your needs?
> 
> In our opinion, a full decoupling is the way to go, vs passing in fused_params to DMP. (We're also rethinking what DMP could look like in the future). I'm going to reject PR for now, but if you think that what we have planned doesn't meet your needs, please comment. cc @YLGH @divchenko

Reiterating Colin, thanks for this PR! IMO, ideally the user is never exposed to fused_params (this naming is unideal). Our design tries maintain that all model behavior is consistent in the unsharded world vs sharded world (so things like optimizer fusion). Whereas tensor placement etc should be done at the sharder level. Actually currently FusedEBC sharder doesn't let you change things like cache_load_factor, but we should definitely support this. Additionally, currently planner doesn't plan around many of these settings (e.g. it assumes you always use rowwise_adagrad, cache_load_factor=20, etc). Moving these settings to sharder + module will (eventually) let planner aware of them.

YLGH:
> > @zhuzilin , thanks for the PR and for the description of the issue you're trying to solve for!
> > We agree, and we have some internal work (and planned work) to address these issues. Firstly, we are working on better decoupling of fusion/batching (i.e compute_kernel in torchrec) and sharding As part of this, we're also making it easier to configure things like optimizer fusion (and optimizer hyperparams) in a more intuitive, pytorch-native way. Like you pointed out, the user can easily forget to pass in optimizer hyperparams and unintentially get different optimizer settings, resulting in bugs.
> > Our proposed solution is to allow the users to better specify the BatchedFused modules (including fused_params) first, and then pass these into DMP (to shard). #505 (just added) includes more information here.
> > Parts of this decomposition are already completed. See https://pytorch.org/blog/pytorch-1.12-new-library-releases/#prototype-single-process-batched--fused-embeddings release notes. You can see this new API (using the fuse_embedding_optimizer wrapper) here: https://github.com/pytorch/torchrec/blob/main/Torchrec_Introduction.ipynb. Can you try this out and see if it meets your needs?
> > In our opinion, a full decoupling is the way to go, vs passing in fused_params to DMP. (We're also rethinking what DMP could look like in the future). I'm going to reject PR for now, but if you think that what we have planned doesn't meet your needs, please comment. cc @YLGH @divchenko
> 
> Reiterating Colin, thanks for this PR! IMO, ideally the user is never exposed to fused_params (this naming is unideal). Our design tries maintain that all model behavior is consistent in the unsharded world vs sharded world (so things like optimizer fusion). Whereas tensor placement etc should be done at the sharder level. Actually currently FusedEBC sharder doesn't let you change things like cache_load_factor, but we should definitely support this. Additionally, currently planner doesn't plan around many of these settings (e.g. it assumes you always use rowwise_adagrad, cache_load_factor=20, etc). Moving these settings to sharder + module will (eventually) let planner aware of them.

Also one outstanding issue that we're in the middle of resolving is if you specify fused_params, or used FusedEBC and specify sharding_type=data_parallel, the embedding tables actually won't have optimizer fusion behavior. We're in the middle of fixing this edge case, but just in case it affects your workflow @zhuzilin

zhuzilin:
@colin2328 @YLGH  Thank you for sharing your ideas and the information on the future plan! I'll close this PR~

And for fused_params, right now we are concerned with optimizer params and cache_load_factor, as the latter would help us train larger models on the same hardware. IMO, they are actually 2 different kinds of params, optimizer params are ones that must be set, otherwise should be deemed as an error, while cache_load_factor is one of the params for intermediate model tweaking. It would be great if the API for them could be in different places :)

> Additionally, currently planner doesn't plan around many of these settings (e.g. it assumes you always use rowwise_adagrad, cache_load_factor=20, etc)

We also noticed this. In fact, "planner ignoring optimizer type" could lead to some wrong sharding partition, e.g. the planner gives a plan that could not run with adam... And for cache_load_factor, we are not sure what's the role of `caching_ratio` in `ParameterConstraints` (I guess it's the same as cache_load_factor :P).

> if you specify fused_params, or used FusedEBC and specify sharding_type=data_parallel, the embedding tables actually won't have optimizer fusion behavior.

This does not affect us at the moment, because we barely have embedding that is small enough to use data_parallel, and even if we do, not fusing them into the whole optimizer should be good (I assume this is only an optimization for performance, not one that will affect correctness?).

YLGH:
>  optimizer params are ones that must be set, otherwise should be deemed as an error,

Totally agree, the behavioral ones that must be set will be through the unsharded module (e.g. explicitly stating the optimizer). other kernel level params (non behavioral cache_load_factor) will be done at the sharder or ParameterPlan level, 
I'll put out a PR for this.

> caching_ratio 

yeah these two are the same :p, i'll put a task in to clean this up. Though one useful thing may be if planner can decide which cache_load_factor is optimal

> assume this is only an optimization for performance, not one that will affect correctness

this can actually affect correctness because currently if you have two separate optimizers one for the model parts, that is optimizer fused, and one for the data_parallel parts, if a table gets moved from model parallel to data_parallel it'll actually affect how it is being optimized

e.g. if you're using rowwise_adagrad for sparse parts, but say torch.optim.SGD over model.parameters(), the training behavior will change if the table gets moved (if data_parallel, it will actually be optimized with torch.optim.SGD). We have some plans in place to fix this (so that even if the model is data_parallel, if it was register under a FusedEBC, it will have the same invariants that after a backwards() pass the model will be automatically updated and gradients will be set to none).

PR #530, opened 2022-07-17
allow override output dtype in quantize_embeddings

zyan0:
Summary: we use activation dtype for QuantEBC output dtype

Differential Revision: D37901080


houseroad (approved):
Looks good.

PR #603, opened 2022-09-05
Copy sparse, dense and labels from the right source when shuffling Criteo data

janekl:
When dense and spare features are located in different directories, i.e., `input_dir_labels_and_dense != input_dir_sparse` as the input to `shuffle()` function, this method throws an error in [L642](https://github.com/pytorch/torchrec/blob/71d51c8764c141ff8d849f73bcf06548ffad36c4/torchrec/datasets/criteo.py#L642) like:
```
FileNotFoundError: [Errno 2] No such file or directory: '/data/contiguous/day_23_dense.npy'
```
So I'm fixing input directories in this PR. This is also in line with [this remark](https://github.com/pytorch/torchrec/blob/71d51c8764c141ff8d849f73bcf06548ffad36c4/torchrec/datasets/scripts/shuffle_preproc_criteo.py#L36-L39).

### DIFF
--- a/torchrec/datasets/criteo.py
+++ b/torchrec/datasets/criteo.py
@@ -632,9 +632,12 @@ class BinaryCriteoUtils:
             curr_first_row = curr_last_row
 
         # Directly copy over the last day's files since they will be used for validation and testing.
-        for part in ["dense", "sparse", "labels"]:
+        for part, input_dir in zip(
+                ["sparse", "dense", "labels"],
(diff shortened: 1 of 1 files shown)


janekl:
cc @samiwilf @narayanan2004

samiwilf:
Thanks @janekl  I hadn't seen this PR until yesterday and had made the fix in a past PR. But the subsidiary changes look good and we'll merge them in.

Issue #528, opened 2022-07-14
TorchRec Sharding Composability


YLGH:
## Desired behavior

We want to make TorchRec sharding composable w/ other sharding/parallelism techniques. This practically means that after applying TorchRec sharding model characteristics remain the same (e.g. state_dict() doesn’t change) and we don’t effect non-sharded parts of a model, e.g.

m = Model(device="meta")
orig_keys = list(m.state_dict().keys())
m.ebc = torchrec.shard_embedding_bag_collection(m.ebc, ....)
sharded_keys = list(m.ebc.state_dict().keys())

assert orig_keys == sharded(keys) # at all ranks, even for table-wise sharded

## Observation for nn.Module

* state_dict() - recursive call via state_dict()
* load_state_dict() : recursive call via local load() call, which calls _load_from_state_dict
* named_parameters()/named_buffers() : calls named_modules
* named_modules() : recursive call

As we don’t want to modify state_dict/named_parameters etc impls of non-sharded modules (including top-level module) we have to ensure that _modules is consistent

## Motivations

* Current implementation (of named_parameters()/named_buffers()/named_modules()) is consistent with nn.module defaults currently
    * This logic is tricky to get right
* Removes need for custom code (e.g tricks in DMP and ShardedModules to keep module FQN consistent)
* Currently state_dict does not contain state of all tensors
    * if a table is sharded table_wise, then only that rank’s state_dict sees this
* easier integration with other distributed solutions (e.g FSDP for DHEN).
    *  We need to share a similar semantics of what is returned by nn.Module APIs, and this will make things consistent.
        * We can also do this within DMP nn.Module overrides, but that route is much heavier and bug-prone
       
## Proposed APIs

Use the trick employed by torch.fx.GraphModule of registering empty nn.Modules for handling state and have implementation of forward() call based on a different module hierarchy.

### new APIs

```
def shard_embedding_modules(model, sharders, plan: EmbeddingShardingPlan):
   # This module will replace all child ebcs/fused_ebcs modules with sharded
   # variants based on the plans and init meta tensors.
   
   # it returns the unsharded modules and their replacements
   # this useful for identifying which modules are already sharded
   return sharded_parameter_names

def dlrm_parallelize(model, 
   sharders=[EmbeddingBagCollectionSharder, FusedEmbeddingBagCollectionSharder, ...], 
   embedding_plan: Optional[EmbeddingShardingPlan]):
  
   # This style of sharding will shard embedding modules with torchrec sharder
[...]

PR #605, opened 2022-09-06
Dynamic Embedding support for TorchRec

zhuzilin:
Hi TorchRec Team!

After a few months of development, our team finished a PoC that adds dynamic embedding support on TorchRec and tested it on our production model. The main idea is to use GPU as the cache for the whole embedding, and store the other part of the embedding on PS. As feed ids are concentrated in most realistic scenarios, such design would make good use of GPU while giving no loss on the precision (AUC) of the model.

## API

As we proposed in the RFC, we add a `wrap` wrapper for the model and dataloader, which would create an asynchronous pipeline for the transformation of the feed ids, fetching of the weights and optimizer states, and eviction internally. An example snippet is:
```python
import torchrec_dynamic_embedding as tde

class Model(nn.Module):
    def __init__(self, config1, config2):
        super().__init__()
        self.emb1 = EmbeddingCollection(tables=config1, device=torch.device("meta"))
        self.emb2 = EmbeddingCollection(tables=config2, device=torch.device("meta"))
        ...

    def forward(self, kjt1, kjt2):
        ...

m = Model(config1, config2)
m = DistributedModelParallel(m)
dataloader = tde.wrap(
    "redis://127.0.0.1:6379/?prefix=model",
    dataloader,
    m,
    # configs of the embedding collections in the model
    { "emb1": config1, "emb2": config2 })

for label, kjt1, kjt2 in dataloader:
    output = m(kjt1, kjt2)
    ...
```

The last dict parameter can be removed if we can access the config from the shared modules.

Apart from `wrap`, we also add a `save` API to flush the GPU cache back to PS.

## ID Transformer
We provide 2 types of `IDTransformer` to transform the real feed ids, whose range would exceed the capacity of GPU (we call them `global_ids`), into the range of GPU caches (we call them `cache_ids`): `NaiveIDTransformer`, which use the `ska::flat_hash_map` to store the mapping and `CachelineTransformer`, which includes a highly customized open address hash table that fully utilizes of the CPU cache line. The performance of the 2 transformers is listed below:

- CPU Info:
  - Run on (16 X 4800 MHz CPU s)
  - CPU Caches:
    - L1 Data 32 KiB (x8)
    - L1 Instruction 32 KiB (x8)
    - L2 Unified 256 KiB (x8)
    - L3 Unified 16384 KiB (x1)
- Performance:
  | type of transformer | num cache embedding | range of global ids | number ids transformed / ms |
  |-|-|-|-|
  | naive |2e8|1e6 ~ 2e6|30.7k|
  | naive |2e8|1e10 ~ 2e10|12.2k|
  | cacheline |2e8|1e6 ~ 2e6|90.6k|
  | cacheline |2e8|1e10 ~ 2e10|23.0k|

The benchmarks could be reproduced with `naive_id_transformer_benchmark.cpp` and `cacheline_id_transformer_benchmark.cpp`.

## Evict Strategy
We provide a `MixedLXUStrategy` for evicting the old ids when the id transformer is full. `MixedLXUStrategy` uses a randomized algorithm similar to [the approximate LFU in redis](https://redis.io/docs/manual/eviction/) to combine both the timestamp and frequency of a global id within 32 bits. The performance of `MixedLXUStrategy` is listed below (CPU is the same as above):
- Update:
  | num of existed ids | num elements to transformer per iter | iteration iter (ms) |
  |-|-|-|
  |3e7|1M|14.7|
  |3e8|1M|14.7|

  (The transform speed is irrelevant to the number of existing ids).
- Eviction:
  | total of ids | num ids to evict | time (s) |
  |-|-|-|
  | 2e7 | 1e7 | 3.5 |
The benchmarks could be reproduced with `mixed_lfu_lru_strategy_benchmark.cpp` and `mixed_lfu_lru_strategy_evict_benchmark.cpp`.

## PS

The current implementation includes a Redis-based PS and a simple C API to register new PS to the library. One could use the `memory_io` in the test as an example to register a new PS. The user **does not** need to add special functionality support (e.g. updating parameter or averaging gradient) when registering their KV system as a PS extension.

## Install and Test

To install the current implementation:
```bash
cd contrib/dynamic_embedding
python setup.py develop
```
And the dynamic embedding will be installed as a separate package named `torchrec_dynamic_embedding`.

We incorporate `gtest` for the C++ code and use `unittest` for the python APIs. The tests make sure that the implementation does not have any precision loss. Please turn on the `TDE_WITH_TESTING` in `setup.py` to run tests. Note that for the python test, one needs to set the environment variable `TDE_MEMORY_IO_PATH` to the path of the compiled `memory_io.so`.

## Experiment on production models

We have tested the current implementation on our production model and observed no AUC loss on both Redis-based PS and simple memory PS. For a model of num embeddings 3e8 (number of ids to hold in embeddings), dynamic embedding enables us to run the model on pure GPU (no UVM involved) by setting the embedding size to 1e7, which improves performance from 7.6 iter/s to 9.0 iter/s.

Even when we still train with num embedding 3e8, the asynchronous pipeline and fast transformers make the performance loss on introducing dynamic embedding neglectable.

We also observed an AUC improvement of 0.3% when changing from `global_id % 3e8` to fit the `int64` global ids into the local machine to using dynamic embedding, a result of preventing any possible id mapping collision (different global id mapped into the same embedding).

Thank you for your time on this PR and we are looking forward to your feedbacks :)

cc @reyoung @colin2328


contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
```diff
@@ -0,0 +1,223 @@
...
+    def _get_data(self):
+        if not self._transform_thread.is_alive():
+            raise RuntimeError("Transform thread exited unexpectedly")
+        data, handles = self._data_queue.get()
+        for handle in handles:
+            handle.wait()
+        return data
+
+
+class DataLoader:
```
colin2328, line 66: is this a dataloader? https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader

see https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
```diff
@@ -0,0 +1,223 @@
...
+    module: DistributedModelParallel,
+    configs_dict: Dict[str, Union[List[EmbeddingBagConfig], List[EmbeddingConfig]]],
+    *,
+    data_info: Dict[int, str] = None,
+    eviction_config=None,
+    transform_config=None,
+    ps_config=None,
+    parallel=True,
+    num_prefetch=0,
+):
```
colin2328, line 141: nit: annotate the return type

colin2328:
API

* we will wrap the dataloader, model and embedding configs? why is this all necessary?
* let’s call the wrapped object a pipeline? (similar to what we do for existing wrapped model/dataloader)? can we use the same progress() API? https://github.com/pytorch/torchrec/blob/main/torchrec/distributed/train_pipeline.py

ID Transformer

* What stage is the ID transformer applied?

Install and Test

* we need to think about this on our end (best way to have this as a submodule)

Experiment on production models

* what is “simple memory” PS?
* “dynamic embedding enables us to run the model on pure GPU (no UVM involved) by setting the embedding size to 1e7, which improves performance from 7.6 iter/s to 9.0 iter/s.”
    * not sure I follow, we are putting 3e8 sized embedding table on 1e7 GPU cache ? 
* ven when we still train with num embedding 3e8, the asynchronous pipeline and fast transformers
    * fast transformers from pytorch? can you share more about model arch?
    * asynchronous pipeline → what are you referring to here?

colin2328 (commented):
initial review with some high level qs

zhuzilin:
@colin2328 
> API
>
> we will wrap the dataloader, model and embedding configs? why is this all necessary?

The id transformer will:
1. transform global id into cache id, i.e. from int64 to the range that we can hold in the local machine.
2. we will need to update the embedding weights when, because we will evict some global ids to PS when the GPU cache is full:
  - when a new global id was mapped into a cache id that was mapped to another global id before eviction. In this case, we need to reinitialized the embedding in the GPU cache, so that the 2 global ids won't share the same embedding or optimizer states;
  - when we find a new global id has embedding stored on PS, we will pull the embedding from PS and copy the value to its new position on the GPU cache.

We need user's dataloader so that we can do 1 and the model so that we can do 2. As far as embedding configs, we use that to get how feature names are mapped to embedding tables. It'll be great if we can extract the config info from the DMP instead of letting the user passes that.

> let’s call the wrapped object a pipeline? (similar to what we do for existing wrapped model/dataloader)? can we use the same progress() API? https://github.com/pytorch/torchrec/blob/main/torchrec/distributed/train_pipeline.py

I think we can.

> ID Transformer
>
> What stage is the ID transformer applied?

A synchronous workflow would be:
1. User's dataloader generates a new batch of data.
2. Id transformer transforms the KJTs in the new batch.
  - 2.1. Id transformer checks if each global id in the KJT is already in the local embedding (has a corresponding cache id).
    - 2.1.1. If there is, then transform the global id to the cache id.
    - 2.1.2 If not, then assign a vacant cache id to this global id.
      - 2.1.2.1. If there is no vacancy, evict half the global ids in the embedding. The ids to evict are determined by our LXU strategy and we will push the embedding of the evicted ids to PS.
      - 2.1.2.2. After eviction, retry transform (2.1).
   - 2.2. Id transformer creates a new KJT those values are cache ids of the global ids.
   - 2.3. Id transformer fetches the embedding of the new global ids from PS into the local embedding.
3. Sync all fetches.
4. User uses the new KJTs(transformed data batch) to do model training.

Note that everything above is done on CPU.

And for an asynchronous workflow, we will add a new python thread for the id transformer:
1. User keeps feeding the new batch into the id transformer thread, which has a fixed-size queue to store the transformed result.
2. User gets the next transformed batch from the queue and synchronizes fetches relate to this batch.
3. User uses the transformed batch to do training.

The async workflow would pipeline the CPU id transform with GPU model training. And that's pretty similar to what `TrainPipelineBase` is doing. So I think we can have an extended version of `TrainPipelineBase` or pipe the id transform pipeline before `TrainPipelineBase`.

> Experiment on production models
> 
> what is “simple memory” PS?

That is referring to the PS in `tests/memory_io`. It's a memory PS that just use `std::unordered_map<std::string, std::vector<uint8_t>>` to store embeddings and we use it to do test.

> “dynamic embedding enables us to run the model on pure GPU (no UVM involved) by setting the embedding size to 1e7, which improves performance from 7.6 iter/s to 9.0 iter/s.”
> not sure I follow, we are putting 3e8 sized embedding table on 1e7 GPU cache ?

The origin model was trained with `global_id % 3e8` and we need to store a 3e8 embedding in torchrec, which needs to use UVM. When testing with local memory PS, we find that can set the embedding size as 1e7 and store the cold embeddings in PS, which make the embedding in torchrec small enough to be held in GPU.

> ven when we still train with num embedding 3e8, the asynchronous pipeline and fast transformers
> fast transformers from pytorch? can you share more about model arch?

Oh, I mean id transformers, sorry for the confusion. The model we tested is a variant of DIN.

> asynchronous pipeline → what are you referring to here?

I'm referring to the async workflow above. It could hide the time consumed by the id transformer, so it won't slow down training.

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
zhuzilin: This is actually a wrapper of the dataloader. And we could turn this into a `TrainPipeline`.

contrib/dynamic_embedding/.clang-format
```diff
@@ -0,0 +1,88 @@
+---
```
divchenko, line 1: @colin2328 should we add one to the top-level /torchrec folder? same as in pytorch?
In general, shall we move this /torchrec/dynamic_embedding folder? This way we can conditionally produce a single python wheel? Just add a command-line flag to setup.py to build dynamic_embedding as well.

contrib/dynamic_embedding/.gitignore
```diff
@@ -0,0 +1,6 @@
+.idea
```
divchenko, line 1: you should be able to just rely on the one from top-level /torchrec folder

contrib/dynamic_embedding/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+cmake_minimum_required(VERSION 3.11.0 FATAL_ERROR)
```
divchenko, line 1: @colin2328 can we make top-level setup.py/CMakeLists ? Then we can just conditionally include contrib/dynamic_embedding from there?

contrib/dynamic_embedding/src/tde/details/url.h
```diff
@@ -0,0 +1,128 @@
+#pragma once
+#include <optional>
+#include "lexy/action/parse.hpp"
+#include "lexy/callback.hpp"
+#include "lexy/dsl.hpp"
+#include "lexy/input/string_input.hpp"
+#include "lexy_ext/report_error.hpp"
```
divchenko, line 7: Including a quite complex third-party to just parse URL seems an overkill.

contrib/dynamic_embedding/src/3rd/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+    # We will not need to test benchmark lib itself.
+    set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "Disable benchmark testing as we don't need it.")
+    # We will not need to install benchmark since we link it statically.
+    set(BENCHMARK_ENABLE_INSTALL OFF CACHE BOOL "Disable benchmark install to avoid overwriting vendor install.")
+    FetchContent_MakeAvailable(googletest google_benchmark)
+endif()
+
+# tcb::span section. A std::span implementation in C++ 11.
+FetchContent_Declare(
```
divchenko, line 23: No need, we just internally moved to C++20, it should be safe to switch in OSS as well?

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+
+    def __len__(self):
+        return len(self._dataloader)
+
+
+def wrap(
+    url: str,
+    dataloader,
+    module: DistributedModelParallel,
```
divchenko, line 132: why specializing to DistributedModelParallel? Can it just be nn.Module?

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
```diff
@@ -0,0 +1,223 @@
...
+                continue
+        # save memory
+        del transformed_data
+
+
+class DataLoaderIter:
+    def __init__(self, dataloader, transform_fn, num_prefetch=0):
+        self._data_queue = queue.Queue(maxsize=num_prefetch)
+        self._done_event = threading.Event()
+        self._transform_thread = threading.Thread(
```
divchenko, line 42: Add a comment that queue is just for prefetch purposes (not parallelization as it's blocked by GIL).

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
```diff
@@ -0,0 +1,223 @@
...
+        return self
+
+    def __next__(self):
+        return self._get_data()
+
+    def __del__(self):
+        self._done_event.set()
+
+    def _get_data(self):
+        if not self._transform_thread.is_alive():
```
divchenko, line 58: this is an odd check

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
```diff
@@ -0,0 +1,223 @@
...
+        module: DMP module that need dynamic embedding.
+        configs_dict: a dictionary that maps the module path of the sharded module to its embedding
+            configs or embeddingbag configs. The plan of `module` should contain the module path
+            in `configs_dict`.
+        data_info: dict keyed by int index of module path. For example, if the dataloader produces
+            `label, kjt1, kjt2` each iteration and `kjt1` and `kjt2` are inputs to modules of path
+            `emb1` and `emb2` respectively, then `data_info` should be `{ 1: "emb1", 2: "emb2" }`.
+        eviction_config: configuration for eviction policy. Default is `{"type": "mixed_lru_lfu"}`
+        transform_config: configuration for the transformer. Default is `{"type": "naive"}`
+        transform_config: configuration for the ps. Default is `{"chunk_size": 8 * 1024 * 1024}
```
divchenko, line 157: ps_config ?

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
```diff
@@ -0,0 +1,223 @@
...
+def wrap(
+    url: str,
+    dataloader,
+    module: DistributedModelParallel,
+    configs_dict: Dict[str, Union[List[EmbeddingBagConfig], List[EmbeddingConfig]]],
+    *,
+    data_info: Dict[int, str] = None,
+    eviction_config=None,
+    transform_config=None,
+    ps_config=None,
```
divchenko, line 138: Use concrete args? i.e. eviction_type, transform_type, etc . It's unclear how to use them w/o reading docs.

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/id_transformer_group.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Check if there is transformer for the path.
+        """
+        return path in self._id_transformer_collections
+
+    def __contains__(self, path):
+        """
+        Check if there is transformer for the path.
+        """
+        return path in self._id_transformer_collections
```
divchenko, line 173: duplicate?

contrib/dynamic_embedding/src/3rd/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+)
+
+# Do not MakeAvailable so do not enable tcb span unittest
+FetchContent_Populate(tcb_span)
+
+add_library(tcb_span INTERFACE)
+target_include_directories(tcb_span INTERFACE ${tcb_span_SOURCE_DIR}/include)
+add_library(tcb::span ALIAS tcb_span)
+
+FetchContent_Declare(
```
divchenko, line 36: can you just pass strongly-typed params to C++ ? it doesn't look like you need a very complex configuration.

contrib/dynamic_embedding/.clang-format
IvanKobzarev, line 1: Lets put `https://github.com/pytorch/pytorch/blob/master/.clang-format` in the root of torchrec and have it as a main .clang-format for all torchrec and remove this file.

.github/workflows/build_dynamic_embedding_wheels.yml
```diff
@@ -0,0 +1,46 @@
+name: Build Dynamic Embedding Wheels
```
IvanKobzarev, line 1: We should have build and test for cpp and py tests here for every PR run

contrib/dynamic_embedding/src/tde/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+# tde_cpp_objs section, used by both unittests and tde_cpp.so
+add_library(tde_cpp_objs OBJECT id_transformer.cpp ps.cpp
+        details/naive_id_transformer.cpp
+        details/cacheline_id_transformer.cpp
+        details/io.cpp details/io_registry.cpp
+        details/move_only_function.cpp
+        details/random_bits_generator.cpp details/mixed_lfu_lru_strategy.cpp
```
IvanKobzarev, line 7: nit: one source file per line please

contrib/dynamic_embedding/src/tde/details/cacheline_id_transformer.h
```diff
@@ -0,0 +1,136 @@
...
+      BitMap,
+      Hash>;
+  static constexpr std::string_view type_ = "cacheline";
+
+  explicit CachelineIDTransformer(int64_t num_embedding, int64_t capacity = 0);
+
+  CachelineIDTransformer(const Self&) = delete;
+  CachelineIDTransformer(Self&&) noexcept = default;
+
+  static Self Create(int64_t num_embedding, const nlohmann::json& json) {
```
IvanKobzarev, line 76: nit: In pytorch we use lowercase methods. Lets make it the same that it will look naturally pytorchy.

contrib/dynamic_embedding/src/tde/details/cacheline_id_transformer.h
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * CachelineIDTransformer
+ *
+ * Transform GlobalID to CacheID by naive flat hash map
+ * @tparam LXURecord The extension type used for eviction strategy.
+ * @tparam Bitmap The bitmap class to record the free cache ids.
+ */
+template <
+    typename LXURecord,
+    int64_t NumCacheline = 8,
```
IvanKobzarev, line 55: why signed types here?

contrib/dynamic_embedding/src/tde/details/notification.h
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Clear the set status.
+   *
+   * NOTE: Clear is not thread-safe.
+   */
+  void Clear();
+
+ private:
+  bool set_{false};
+  std::mutex mtx_;
```
IvanKobzarev, line 24: please add comment what is it guarding

contrib/dynamic_embedding/src/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(3rd)
```
IvanKobzarev, line 1: Lets specify which C++ version is it, C++ 17,
set(CMAKE_CXX_STANDARD 17 CACHE STRING "The C++ standard whose features are requested to build this target.")

IvanKobzarev (commented):
Lets separate cpp and python code as we do in pytorch/pytorch
Sources:
dynamic_embedding/csrc/**.h **.cpp
dynamic_embedding/*.py

Tests:
dynamic_embedding/test/*.py
dynamic_embedding/test/cpp/*.py

contrib/dynamic_embedding/CMakeLists.txt
colin2328: we can, it depends on how much we want to separate these features out. Let's think more about this but this is fine for now

colin2328 (approved):
Thanks for the feature! As discussed, let's land this now (we will do it for you) and we will give more detailed reviews as you move from existing contrib into torchrec. We  will follow up with proposed structure

split out into multiple diffs
1) the high level API
2) Id Transformer
3) parameter server
4) eviction strategies

High level suggestions:
lets create DynamicEmbeddingCollection, DynamicEmbeddingBagCollection and
ShardedDynamicEmbeddingCollectoin and ShardedDynamicEmbeddingBagCollection. The sharded versions will be very similiar to non dynamic (subclass)

let’s move tde.wrap() to create a train_pipeline (with progress() API). prefetching will be one of the stages

We will keep adding comments/suggestions here in the meantime. We’ll also put out some proposed interfaces

contrib/dynamic_embedding/src/3rd/CMakeLists.txt
reyoung: Sure. The IDTransformer configuration should be strongly typed.

We used `JSON` before because the `TorchRec` training is too fast, and the IDTransformer could be a bottleneck in the system. So we actually tried many kinds of IDTransformer implementation before. 

Since the number of kinds is limited now, we can change the configuration to strongly typed now.

YLGH:
Sorry for being late to the discussion, finally wrapping my head around everything :p

One QQ - that I was discussing with @colin2328,

Would it be possible for this new model type DynamicEmbeddingBagCollection to take in global_id_keys, rather than local? It seems more natural to use (easier to think about, and closer to original EBCs)

underneath, the prefetch logic could be done as a separate stage in the model, prior to any compute, so I imagine on the unsharded module this would look like:

1. prefetch
2. compute

and sharded module would look like

1. prefetch
2. input_dist
3. compute
4. output_dist

If you use this by default, then you could block on some iterations, but a clever training pipeline (as colin suggested) we can do some fx + reording magic to overlap this with computation 

Also, this may be a moot point with the new DynamicModules, but we want to move away from DistributedModelParallel in the future - and have started some work on this path, see https://github.com/pytorch/torchrec/issues/528 for some more details, hopefully we can remove some of the logic of module path traversals

YLGH:
> Sorry for being late to the discussion, finally wrapping my head around everything :p
> 
> One QQ - that I was discussing with @colin2328,
> 
> Would it be possible for this new model type DynamicEmbeddingBagCollection to take in global_id_keys, rather than local? It seems more natural to use (easier to think about, and closer to original EBCs)
> 
> underneath, the prefetch logic could be done as a separate stage in the model, prior to any compute, so I imagine on the unsharded module this would look like:
> 
> 1. prefetch
> 2. compute
> 
> and sharded module would look like
> 
> 1. prefetch
> 2. input_dist
> 3. compute
> 4. output_dist
> 
> If you use this by default, then you could block on some iterations, but a clever training pipeline (as colin suggested) we can do some fx + reording magic to overlap this with computation
> 
> Also, this may be a moot point with the new DynamicModules, but we want to move away from DistributedModelParallel in the future - and have started some work on this path, see #528 for some more details, hopefully we can remove some of the logic of module path traversals

not sure if folks will see if this issue is closed, so just retagging @zhuzilin

contrib/dynamic_embedding/src/torchrec_dynamic_embedding/dataloader.py
colin2328: yes, we think this is important to make this compatible with the composable design (shifting from DMP to shard_embedding_modules) see https://github.com/pytorch/torchrec/issues/528

zhuzilin:
@YLGH Thank you for sharing! We are open to all possibilities to the API :) 

After the last meeting, we recognized that having a DynamicEBC is a really good idea, which makes the dynamic embedding more coherent to the current torchrec codebase. And we agree that `TrainingPipeline` is an ideal place to do prefetch.
As far as `torch.fx`, we are not familiar with it now and may need some time to learn.

Also, we'd love to migrate the C++ part, which should be more stable, to top level folder, while discussing API with you.

backyes:
@zhuzilin What's the progress of dynamic  embedding support for torchrec? It seems to have no active dissussion on this topic. 

Thanks.

zhuzilin:
@backyes We have migrated and cleaned all the c++ code into [`torchrec/csrc`](https://github.com/pytorch/torchrec/tree/main/torchrec/csrc) folder. And I believe the torchrec team is working on creating a better python api for it.

PR #661, opened 2022-09-26
[tde] Migrate naive id transformer

zhuzilin:
This PR is trying to migrate `NaiveIDTransformer` to the top folder. The idea behind the naive transformer is quite simple, we use a hash map (`ska::flat_hash_map`) to store the global id to cache id map, and use a bitmap to record the empty slot in the cache.

cc @reyoung


zhuzilin:
gently ping @colin2328 @divchenko

colin2328:
looks fine to me. @IvanKobzarev any thoughts on style?

colin2328:
@zhuzilin  -> which object will use NaiveIDTransformer (python wrapper) as state? will it live on each trainer? (and each trainer will have the mapping?) cc @divchenko

zhuzilin:
> @zhuzilin -> which object will use NaiveIDTransformer (python wrapper) as state? will it live on each trainer? (and each trainer will have the mapping?) cc @divchenko

Currently we only need an id transformer on rank 0, and global ids on all ranks will be gathered to rank 0.

colin2328:
@zhuzilin  I see, and all ofthis will be done in the prefetching stage of the train_pipeline?

zhuzilin:
> @zhuzilin I see, and all ofthis will be done in the prefetching stage of the train_pipeline?

yes, exactly.

colin2328 (approved):
LGTM thanks

PR #662, opened 2022-09-26
Direct planner

YLGH:
Summary:
See  https://docs.google.com/document/d/1ASiz0Fu3c8cCHmsbseUdulU4jfTgdwZAxvlywgXiBAU/edit for more context

tl;dr:

Currently the only “feasible” way to specify a sharding plan is by using the OSS planner, and using constraints. I say “feasible” because although we can manually construct a ShardingPlan/ParameterSharding, it’s quite cumbersome to do so. Additionally, users do not have the expressibility to say “I want to place this parameter on rank 0” when using planner.

With this new API we can construct sharding plans in a direct way:

```
ebc = EmbeddingBagCollection(...)
plan = construct_parameter_sharding_plans(
            ebc,
            {
                "table_0": data_parallel_sharding(),
                "table_1": row_wise_sharding(),
                "table_2": column_wise_sharding(),
                "table_3": column_wise_sharding(ranks=[0,1,2]),
                "table_4": table_row_wise_sharding(host_index=2),
            },
            EmbeddingBagCollectionSharder()
        )
ebc = sharder.shard(
   ebc,
   plan
)
```

Note, that this is not intended to replace the greedy planner. Instead it allows users to shard modules without being FORCED to use planner. Users will be able to user planner in the exact same way.

Having alternatives instead of planner has a couple of benefits

1. There are bugs in the planner (we should be able to shard something, but planner may have a false negative rejection)
2. Planner doesn't correctly account for new architectures/models /optimizers
3. Easier to directly use torchrec (new users don't need to understand how to use planner).

-----

So I decided to use a generator function because alternative ways would require creating a new dataclass called like TableWiseConfig(row=...), and I didn't like the look of it.

Having just the atomic generator functions is also nice since it doesn't couple these configs with the construct_parameter_sharding_plans

Differential Revision: D39803183

### DIFF
--- a/torchrec/distributed/planner/planners.py
+++ b/torchrec/distributed/planner/planners.py
@@ -14,6 +14,7 @@ import torch
 import torch.distributed as dist
 from torch import nn
 from torchrec.distributed.collective_utils import invoke_on_rank_and_broadcast_result
+from torchrec.distributed.parameter_sharding_utils import placement
 from torchrec.distributed.planner.constants import BATCH_SIZE, MAX_SIZE
 from torchrec.distributed.planner.enumerators import EmbeddingEnumerator
 from torchrec.distributed.planner.partitioners import GreedyPerfPartitioner
@@ -56,15 +57,6 @@ def _to_sharding_plan(
     sharding_options: List[ShardingOption],
     topology: Topology,
 ) -> ShardingPlan:
-    def _placement(
-        compute_device: str,
-        rank: int,
-        local_size: int,
-    ) -> str:
-        param_device = compute_device
-        if compute_device == "cuda":
-            param_device = torch.device("cuda", rank % local_size)
-        return f"rank:{rank}/{param_device}"
 
     compute_device = topology.compute_device
     local_size = topology.local_world_size
@@ -83,7 +75,7 @@ def _to_sharding_plan(
                     ShardMetadata(
                         shard_sizes=shard.size,
                         shard_offsets=shard.offset,
-                        placement=_placement(
+                        placement=placement(
                             compute_device, cast(int, shard.rank), local_size
                         ),
                     )
--- a/torchrec/distributed/model_parallel.py
+++ b/torchrec/distributed/model_parallel.py
@@ -18,16 +18,13 @@ from torch.distributed.fsdp import FullyShardedDataParallel
 from torch.nn.modules.module import _IncompatibleKeys
 from torch.nn.parallel import DistributedDataParallel
 from torchrec.distributed.comm import get_local_size
-from torchrec.distributed.embedding import EmbeddingCollectionSharder
-from torchrec.distributed.embeddingbag import EmbeddingBagCollectionSharder
-from torchrec.distributed.fused_embeddingbag import FusedEmbeddingBagCollectionSharder
+
+from torchrec.distributed.parameter_sharding_utils import get_default_sharders
 from torchrec.distributed.planner import (
     EmbeddingShardingPlanner,
     sharder_name,
     Topology,
 )
-from torchrec.distributed.quant_embedding import QuantEmbeddingCollectionSharder
-from torchrec.distributed.quant_embeddingbag import QuantEmbeddingBagCollectionSharder
 from torchrec.distributed.types import (
     ModuleCopyMixin,
     ModuleSharder,
@@ -46,16 +43,6 @@ from torchrec.optim.keyed import CombinedOptimizer, KeyedOptimizer
 
 
 _DDP_STATE_DICT_PREFIX = "module."
-
-
-def get_default_sharders() -> List[ModuleSharder[nn.Module]]:
-    return [
-        cast(ModuleSharder[nn.Module], EmbeddingBagCollectionSharder()),
-        cast(ModuleSharder[nn.Module], FusedEmbeddingBagCollectionSharder()),
-        cast(ModuleSharder[nn.Module], EmbeddingCollectionSharder()),
-        cast(ModuleSharder[nn.Module], QuantEmbeddingBagCollectionSharder()),
-        cast(ModuleSharder[nn.Module], QuantEmbeddingCollectionSharder()),
-    ]
 
 
 class DataParallelWrapper(abc.ABC):
--- a/torchrec/distributed/parameter_sharding_utils.py
+++ b/torchrec/distributed/parameter_sharding_utils.py
@@ -1,0 +1,513 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+#!/usr/bin/env python3
+
+import math
+from typing import Callable, cast, Dict, List, Optional, Tuple, Type
+
+import torch
+from torch import distributed as dist, nn
+from torchrec.distributed.comm import get_local_size
+from torchrec.distributed.embedding import EmbeddingCollectionSharder
+from torchrec.distributed.embedding_types import EmbeddingComputeKernel
+from torchrec.distributed.embeddingbag import EmbeddingBagCollectionSharder
+from torchrec.distributed.fused_embeddingbag import FusedEmbeddingBagCollectionSharder
+from torchrec.distributed.quant_embedding import QuantEmbeddingCollectionSharder
+from torchrec.distributed.quant_embeddingbag import QuantEmbeddingBagCollectionSharder
+from torchrec.distributed.types import (
+    EnumerableShardingSpec,
+    ModuleSharder,
+    ParameterSharding,
+    ShardingType,
+    ShardMetadata,
+)
+from torchrec.distributed.utils import none_throws
+
+MIN_CW_DIM: int = 128
+
+
+def get_default_sharders() -> List[ModuleSharder[nn.Module]]:
+    return [
+        cast(ModuleSharder[nn.Module], EmbeddingBagCollectionSharder()),
+        cast(ModuleSharder[nn.Module], EmbeddingCollectionSharder()),
+        cast(ModuleSharder[nn.Module], FusedEmbeddingBagCollectionSharder()),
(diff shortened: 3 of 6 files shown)


torchrec/distributed/planner/tests/test_simple_planner.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def _test_sharding(
+    tables: List[EmbeddingBagConfig],
+    initial_state_dict: Dict[str, Any],
+    rank: int,
+    world_size: int,
+    kjt_input_per_rank: List[KeyedJaggedTensor],
+    sharder: ModuleSharder[nn.Module],
+    backend: str,
```
divchenko: nit: group rank, world_size, local size, backend together . Maybe introduce a dataclass so it's easier to pass all the way to MultiProcessContext ?

torchrec/distributed/planner/tests/test_simple_planner.py
```diff
@@ -0,0 +1,232 @@
...
+
+        feature_keys = []
+        for table in tables:
+            feature_keys.extend(table.feature_names)
+
+        # each rank gets a subbatch
+        sharded_model_pred_kt = sharded_model(kjt_input_per_rank[ctx.rank]).to_dict()
+        _sharded_model_pred = torch.stack(  # noqa
+            [sharded_model_pred_kt[feature] for feature in feature_keys]
+        )
```
divchenko: Do you intend to check something here? e.g. sharded vs unsharded results?

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+
+    per_parameter_sharding: Dict[str, ParameterSharding] = {}
+
+    for param_name, sharding_plan_generator in per_param_sharding.items():
+        param = None
+        for name, _param in module.named_parameters():
+            if param_name in name:
+                param = _param
+                break
+        assert param is not None
```
divchenko: better error message

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+        )
+
+    return _parameter_sharding_generator
+
+
+def construct_parameter_sharding_plans(
+    module: nn.Module,
+    per_param_sharding: Dict[
+        str, Callable[[nn.Parameter, int, int, str], ParameterSharding]
+    ],
```
divchenko: This Callable is complex and repeated in many places. Define a new type.
ParameterShardingGenerator = Callable[[nn.Parameter, int, int, str], ParameterSharding]

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+        assert param is not None
+
+        per_parameter_sharding[param_name] = sharding_plan_generator(
+            param, local_size, world_size, device_type
+        )
+
+    return per_parameter_sharding
+
+
+def uniform_sharding(
```
divchenko: What is the final end-to-end code? We need to list an example somewhere.
```
model = MyModel()
ebc_plan = construct_parameter_sharding_plans(model.ebc, uniform_sharding(model.ebc, row_wise_sharding())
eb_plan = construct_parameter_sharding_plans(model.eb, uniform_sharding(model.eb, row_wise_sharding())
plan = {**eb_plan, **eb_plan}
```

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+
+    return _parameter_sharding_generator
+
+
+def construct_parameter_sharding_plans(
+    module: nn.Module,
+    per_param_sharding: Dict[
+        str, Callable[[nn.Parameter, int, int, str], ParameterSharding]
+    ],
+    sharder: ModuleSharder[nn.Module],
```
divchenko: We need to add a higher-level API, where we don't pass sharder. There is a module -> sharder dict. It should operate on top of it.

divchenko (approved):

torchrec/distributed/planner/tests/test_simple_planner.py
YLGH: here I just wanted to check that things run w/ the constructed plans,

there will be another unit test that does correctness validation over composable ebc with inputs from this

torchrec/distributed/planner/planners.py
YLGH: the param_name in name is also a bit annoying because we strip off ".weight", I guess I can get substring bugs

torchrec/distributed/planner/planners.py
YLGH: yeah 
I think there needs to be something that can add FQN path prefix to output of construct_parameter_sharding_plans: Dict[str, ParameterSharding], 

this operates on the Sharder.shard level

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+
+    return per_parameter_sharding
+
+
+def uniform_sharding(
+    module: nn.Module,
+    parameter_sharding_generator: Callable[
+        [nn.Parameter, int, int, str], ParameterSharding
+    ],
+    sharder: ModuleSharder[nn.Module],
```
YLGH: default to None and try to find the correct one from dict

torchrec/distributed/planner/planners.py
YLGH:
```
Overall plan would look like
{
   "model.ebc": ebc_plan,
   "model.ec" : ec_plan
}
``` i suppose

File: torchrec/distributed/planner/planners.py [hunk 8]
```diff
@@ -302,3 +300,267 @@ def plan(
...
+                        none_throws(shard.rank),
+                        none_throws(local_size),
+                    ),
+                )
+                for shard in shards
+            ]
+        ),
+        sharding_type=sharding_type,
+        # pyre-ignore
+        compute_kernel=None,
```
colin2328: setting compute_kernel to None will mean that we cannot use w/ DMP (unless manually changed), IMO add comment here or in user facing API

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+
+        return _get_parameter_sharding(
+            ShardingType.ROW_WISE.value, shards, local_size, device_type
+        )
+
+    return _parameter_sharding_generator
+
+
+def column_wise_sharding(
+    shard_dim: int = 128,
```
colin2328: add docstrings, including what shard_dim represents

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+) -> Dict[str, ParameterSharding]:
+    """
+    Helper function to create sharding plans (Dict[str, ParameterSharding]) for an module
+    Args:
+        module (nn.Module): module to create plan for.
+        per_param_sharding: Dict[str, Callable[[nn.Parameter, int, int, str], ParameterSharding]]: A mapping of parameter names to a generator function
+        that takes in [parameter, local_size, world_size, device_type] and returns a ParameterSharding. We recommend using one of the predefined generator functions
+        e.g. table_wise_sharding, row_wise_sharding, etc,
+        sharder: ModuleSharder[nn.Module]: Sharder that we are creating a plan for,
+        local_size: Optional[int] = None: process group local size
```
colin2328: add generally user should not pass these in

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+        world_size: Optional[int] = None: process_group world_size
+        device_type: str : Torch device type,
```
colin2328: IMO we should use device, or at least torch.device.type rather than str

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+
+    Example::
+        ebc = EmbeddingBagCollection(...)
+        plan = construct_parameter_sharding_plan(
+            ebc,
+            {
+                "table_0": data_parallel_sharding(),
+                "table_1": row_wise_sharding(),
+                "table_2": column_wise_sharding(),
+                "table_3": column_wise_sharding(ranks=[0,1,2]),
```
colin2328: consider allowing this, or if not at least update to reflect (need to specify shard_dim)

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+                "table_4": table_row_wise_sharding(host_index=2),
```
colin2328: we need to write it or comment that it's not supported

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+
+    local_size = local_size or get_local_size()
+    world_size = world_size or dist.get_world_size()
+
+    per_parameter_sharding: Dict[str, ParameterSharding] = {}
+
+    for param_name, sharding_plan_generator in per_param_sharding.items():
+        param = None
+        for name, _param in module.named_parameters():
+            if param_name in name:
```
colin2328: we need to strip weight and actually match rather than relying on substring match

colin2328: since you already do assertion above just iterate through and do .get()

torchrec/distributed/planner/planners.py
```diff
@@ -302,3 +300,267 @@ def plan(
...
+    ],
+    sharder: ModuleSharder[nn.Module],
+) -> Dict[str, Callable[[nn.Parameter, int, int, str], ParameterSharding]]:
+    """
+    Convenience function to generate a uniform sharding scheme for construct_parameter_sharding_plans.
+
+    Example::
+    ebc = EmbeddingBagCollection(...)
+    sharder = EmbeddingBagCollectionSharder()
+    plan = construct_parameter_sharding_plan(
```
colin2328: IMO we should avoid uniform and instead call it something that indicates were sharding all tables like this "shard_all_tables"

colin2328:
nice

PR #678, opened 2022-09-30
Decouple compute kernel from sharding

YLGH:
Summary:
This is a fully backwards compatible change to allow users to have model authoring code that's decoupled compute_kernels from sharding.

If compute kernel is set, this will be fully backwards compatible

We can derive what compute kernel we should use entirely based on the parameter's attributes as well as sharding type: e.g.

1. If data_parallel, always use dense kernel
2. If model_parallel and has overlapped_optimizer (that's supported), use FUSED
3. If no overlapped optimizer, use dense

Other subtely is that we still want to enable placement. We create a couple new classes, the important one is UVM_CACHING, which has a field called caching ratio

The fully authoring experience could then look like

```
ebc = EmbeddingBagCollection(...)
plan = construct_parameter_sharding_plans(
            ebc,
            {
                "table_0": data_parallel_sharding(),
                "table_1": row_wise_sharding(placement=UVM_CACHING(caching_ratio=0.2)),
                "table_2": column_wise_sharding(placement=UVM),
                "table_3": column_wise_sharding(ranks=[0,1,2]),
                "table_4": table_row_wise_sharding(host_index=2),
            },
        )
ebc = sharder.shard(
   ebc,
   plan
)
```

Note, this doesn't work with planner yet, but should be easy to incorporate.

Differential Revision: D39975761

### DIFF
--- a/torchrec/distributed/embeddingbag.py
+++ b/torchrec/distributed/embeddingbag.py
@@ -34,6 +34,7 @@ from torchrec.distributed.types import (
     Awaitable,
     EmptyShardedModuleContext,
     EnumerableShardingSpec,
+    HBM,
     LazyAwaitable,
     ParameterSharding,
     QuantizedCommCodecs,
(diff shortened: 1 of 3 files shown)


torchrec/distributed/planner/planners.py
```diff
@@ -7,14 +7,19 @@
```
YLGH: ignore the changes here,

divchenko:
Great idea!
In future we will need to change automated sharder to remove handling of compute kernels there.

PR #730, opened 2022-10-17
[tde] Migrate IO plugin

zhuzilin:
This PR implements the IO plugin mechanism in dynamic embedding, where user can register her/his IO plugin by providing a dynamic library with symbol `IO_type`, `IO_Initialize`, `IO_Finalize`, `IO_Push` and `IO_Pull` included.

And the tests for the interfaces introduced in this PR will be added along with a example redis IO plugin in next PR. I split them so that the PR won't be too large.

Thank you for your time on reviewing this PR :)

cc @reyoung @colin2328

### DIFF
--- a/torchrec/csrc/dynamic_embedding/details/io.cpp
+++ b/torchrec/csrc/dynamic_embedding/details/io.cpp
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <torchrec/csrc/dynamic_embedding/details/io.h>
+
+namespace tde::details {
+
+static constexpr std::string_view k_schema_separator = "://";
+
+IO::IO(const std::string& config) {
+  auto pos = config.find(k_schema_separator);
+  TORCH_CHECK(
+      pos != std::string::npos,
+      "config string should be schema://cfg_string, cannot find schema");
+
+  std::string schema = config.substr(0, pos);
+  std::string rest_cfg = config.substr(pos + k_schema_separator.size());
+  auto& reg = IORegistry::Instance();
+  provider_ = reg.resolve(schema);
+  instance_ = provider_.initialize_(rest_cfg.c_str());
+}
+
+IO::~IO() {
+  if (instance_ == nullptr) {
+    return;
+  }
+  provider_.finalize_(instance_);
+}
+
+struct FetchContext {
+  std::vector<torch::Tensor> tensors_;
+  std::function<void(std::vector<torch::Tensor>)> on_complete_;
+  torch::ScalarType scalar_type_;
+  uint32_t num_optimizer_states_;
+};
+
+static void on_global_id_fetched(
+    void* ctx,
+    uint32_t offset,
+    uint32_t optimizer_state,
+    void* data,
+    uint32_t data_len) {
+  auto c = reinterpret_cast<FetchContext*>(ctx);
+  auto& tensor = c->tensors_[offset];
+  // non-existed global id
+  if (data_len == 0) {
+    if (tensor.defined()) {
+      tensor = torch::Tensor{};
+    }
+    return;
+  }
+  if (!tensor.defined()) {
+    size_t elem_size = torch::elementSize(c->scalar_type_);
+    TORCH_CHECK(data_len % elem_size == 0);
+    size_t num_elems = data_len / elem_size;
+    tensor = torch::empty(
+        {
+            static_cast<int64_t>(c->num_optimizer_states_),
+            static_cast<int64_t>(num_elems),
+        },
+        c10::TensorOptions().dtype(c->scalar_type_));
+  }
+  void* ptr = reinterpret_cast<void*>(
+      reinterpret_cast<uintptr_t>(tensor.data_ptr()) +
+      optimizer_state * data_len);
+  memcpy(ptr, data, data_len);
+}
+
+static void on_all_fetched(void* ctx) {
+  auto c = reinterpret_cast<FetchContext*>(ctx);
+  c->on_complete_(std::move(c->tensors_));
+  delete c;
+}
+
+void IO::pull(
+    const std::string& table_name,
+    std::span<const int64_t> global_ids,
+    std::span<const int64_t> col_ids,
+    uint32_t num_optimizer_states,
+    torch::ScalarType type,
+    std::function<void(std::vector<torch::Tensor>)> on_fetch_complete) {
+  std::unique_ptr<FetchContext> ctx(new FetchContext{
+      .on_complete_ = std::move(on_fetch_complete),
+      .scalar_type_ = type,
+      .num_optimizer_states_ = num_optimizer_states,
+  });
+
+  ctx->tensors_.resize(
+      global_ids.size() * std::max(col_ids.size(), static_cast<size_t>(1)));
+
+  IOPullParameter param{
+      .table_name_ = table_name.c_str(),
+      .num_cols_ = static_cast<uint32_t>(col_ids.size()),
+      .num_global_ids_ = static_cast<uint32_t>(global_ids.size()),
+      .col_ids_ = col_ids.data(),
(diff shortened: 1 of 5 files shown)


torchrec/csrc/dynamic_embedding/details/io.cpp
```diff
@@ -0,0 +1,148 @@
...
+    return;
+  }
+  provider_.finalize_(instance_);
+}
+
+struct FetchContext {
+  std::vector<torch::Tensor> tensors_;
+  std::function<void(std::vector<torch::Tensor>)> on_complete_;
+  torch::ScalarType scalar_type_;
+  uint32_t num_optimizer_states_;
```
divchenko: these are all public memebers, no need to suffix names w/ "_"

torchrec/csrc/dynamic_embedding/details/io.h
```diff
@@ -0,0 +1,78 @@
...
+      std::span<const int64_t> global_ids,
+      std::span<const int64_t> col_ids,
+      uint32_t num_optimizer_states,
+      torch::ScalarType type,
+      std::function<void(std::vector<torch::Tensor>)> on_fetch_complete);
+
+  /**
+   * Push Parameter/Optimizer stats to parameter server.
+   * @param table_name
+   * @param global_ids
```
divchenko: omit params for which you don't provide extra info

torchrec/csrc/dynamic_embedding/details/io_registry.h
```diff
@@ -0,0 +1,84 @@
...
+  const int64_t* global_ids_;
+  uint32_t num_optimizer_stats_;
+  void* on_complete_context_;
+  void (*on_global_id_fetched_)(
+      void* ctx,
+      uint32_t offset,
+      uint32_t optimizer_state,
+      void* data,
+      uint32_t data_len);
+  void (*on_all_fetched_)(void* ctx);
```
divchenko: Here and everywhere else: these are all public members, no need to suffix names w/ "_"

torchrec/csrc/dynamic_embedding/details/io_registry.h
```diff
@@ -0,0 +1,84 @@
...
+  uint32_t num_optimizer_stats_;
+  const uint32_t* optimizer_stats_ids_;
+  // offsets in bytes
+  // data_ptr is 1-D array divided by offsets in bytes.
+  // The offsets are jagged and length of offsets array is length + 1.
+  //
+  // data[global_id * num_cols * num_optimizer_stats_
+  //       + col_id * num_optimizer_stats_ + os_id ]
+  uint32_t num_offsets_;
+  const uint64_t* offsets_;
```
divchenko: combine into std::span?

torchrec/csrc/dynamic_embedding/details/io.h
```diff
@@ -0,0 +1,78 @@
...
+   * different in some algorithm.
+   *
+   * @note this method is asynchronous, The col_ids, and global_ids will be
+   * copied inside, so it is safe to free col_ids/global_ids before
+   * on_fetch_complete.
+   */
+  void pull(
+      const std::string& table_name,
+      std::span<const int64_t> global_ids,
+      std::span<const int64_t> col_ids,
```
divchenko, line 50: Explain in comments what col_ids is.

torchrec/csrc/dynamic_embedding/details/io.h
```diff
@@ -0,0 +1,78 @@
...
+   * os_id_offset]
+   *
+   * @param offsets
+   * @param on_push_complete
+   */
+  void push(
+      const std::string& table_name,
+      std::span<const int64_t> global_ids,
+      std::span<const int64_t> col_ids,
+      std::span<const uint32_t> os_ids,
```
divchenko, line 82: Explain in comments what os_ids is

torchrec/csrc/dynamic_embedding/details/io.h
```diff
@@ -0,0 +1,78 @@
...
+   * @param col_ids empty if no column slices
+   * @param os_ids
+   * @param data A flatten view of pushing data.
+   * data[gid_offset * num_cols * num_os_id + col_offset * num_os_id +
+   * os_id_offset]
+   *
+   * @param offsets
+   * @param on_push_complete
+   */
+  void push(
```
divchenko, line 78: This API is a bit asymmetric. You pull Tensor , but you push raw data (i.e data + offsets).

torchrec/csrc/dynamic_embedding/details/io_registry.h
```diff
@@ -0,0 +1,84 @@
...
+
+namespace tde::details {
+
+struct IOPullParameter {
+  const char* table_name_;
+  uint32_t num_cols_;
+  uint32_t num_global_ids_;
+  const int64_t* col_ids_;
+  const int64_t* global_ids_;
+  uint32_t num_optimizer_stats_;
```
divchenko: num_optimizer_states?

torchrec/csrc/dynamic_embedding/details/io_registry.h
```diff
@@ -0,0 +1,84 @@
...
+  void (*on_all_fetched_)(void* ctx);
+};
+
+struct IOPushParameter {
+  const char* table_name_;
+  uint32_t num_cols_;
+  uint32_t num_global_ids_;
+  const int64_t* col_ids_;
+  const int64_t* global_ids_;
+  uint32_t num_optimizer_stats_;
```
divchenko: num_optimizer_states?

divchenko (approved):

torchrec/csrc/dynamic_embedding/details/io.cpp
zhuzilin: fixed.

torchrec/csrc/dynamic_embedding/details/io.h
zhuzilin: fixed.

torchrec/csrc/dynamic_embedding/details/io_registry.h
zhuzilin: fixed.

torchrec/csrc/dynamic_embedding/details/io.h
zhuzilin: Added. `col_ids` is the column id of the embedding. Because an embedding will be sharded with column-wise sharding, we hope to store different segments as different kv pairs in PS. And the column id represents the index of the column segment.

torchrec/csrc/dynamic_embedding/details/io.h
zhuzilin: Added. `os_ids` is the id of the optimizer states. For example, we need to not only store the parameter, but also `m` and `v` in the optimizer, therefore, we use `os_id` to differentiate them.

torchrec/csrc/dynamic_embedding/details/io.h
zhuzilin: Yeah, it's asynmmetric here, the reason is that we hope to push data by chunks and hope to share the buffer between chunks. We have grouped the data into chunks outside this api, where we transform the tensor into  chunks. Maybe we can come back to this api when migrating the outer api?

torchrec/csrc/dynamic_embedding/details/io_registry.h
zhuzilin: fixed.

torchrec/csrc/dynamic_embedding/details/io_registry.h
zhuzilin: fixed.

torchrec/csrc/dynamic_embedding/details/io_registry.h
zhuzilin: Here we were hoping to export an C api, so that user could write the plugin in whatever language they prefer. And that's the reason why we split the `std::span` into a pointer and its size and also split `std::function` into a function pointer and its context.

zhuzilin:
@divchenko Sorry for this late reply, we were experimenting with our inner PS to make sure the current plugin design could work with it. I have updated the code upon your review, could you take another look? Thank you!

colin2328 (approved):

PR #775, opened 2022-11-02
[Docs] Fix `ComputeKJTToJTDict` docstring render

awgu:
I am not sure if CI renders docs as a part of testing. This is not meant for landing until doc render has been verified.

Before change:
Screen Shot 2022-11-02 at 1 41 13 PM


awgu:
It looks like this did not actually render correctly. I am not sure how to build the docs manually myself, but pattern matching against the ones that do render correctly, I think the mistake with this PR is that I did not include a newline after `Example::` and before the actual example contents. Let me follow up.

PR #790, opened 2022-11-07
Add support for materializing and reading materialized 1tb criteo multi-hot dataset

samiwilf:
Differential Revision: D41086252


colin2328 (changes requested):
IMO we don't want to add this to torchrec/datasets/criteo. 
this (multihot dataset support) is very specific to DLRMV2, and should live in DLRMv2 repo . Feel free to reuse any of the dataset utilitiies from torchrec

PR #835, opened 2022-11-21
[tde] Migrate id transformer binding

zhuzilin:
This PR integrated the migrated id transformer and lxu strategy to create a python binding. There are several point that may need to be mentioned:
- We use `IDTransformerVariant` and `LXUStrategyVariant` type to support adding more kinds of id transformers and lxu strategies.
- The origin design used [nlohmann::json](https://github.com/nlohmann/json) to parse the configs. To remove this dependency, we list all configs of all kinds of id transformers and lxu strategies as the parameter of the constructor of the variant.
- We choose to use pytorch native binding (`CustomClassHolder`) instead of pybind to be coherent with fbgemm_gpu.

Thank you for your time on reviewing this PR :)

gently ping @divchenko @colin2328 @reyoung


torchrec/csrc/dynamic_embedding/tensor_list.h
```diff
@@ -0,0 +1,41 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+#include <torch/torch.h>
+
+namespace torchrec {
+
+class TensorList : public torch::CustomClassHolder {
```
divchenko: Looks like it's only used to pass a list of tensors from Python -> C++ . You should be able to pass a Python list of Tensors via pybind and then access them via pybind API.

torchrec/csrc/dynamic_embedding/details/id_transformer_variant.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace torchrec {
+
+IDTransformerVariant::IDTransformerVariant(
+    LXUStrategyVariant strategy,
+    int64_t num_embeddings,
+    const std::string& id_transformer_type)
+    : strategy_(std::move(strategy)),
+      var_(NaiveIDTransformer<uint32_t>(num_embeddings)) {
+  if (id_transformer_type != "naive") {
```
divchenko: This check should not be done here (should be somewhere in the upper levels). Thus you don't need to pass id_transformer_type at all.

torchrec/csrc/dynamic_embedding/details/id_transformer_variant.cpp
```diff
@@ -0,0 +1,80 @@
...
+        return result;
+      },
+      var_);
+}
+
+LXUStrategyVariant::LXUStrategyVariant(
+    const std::string& lxu_strategy_type,
+    uint16_t min_used_freq_power)
+    : strategy_(MixedLFULRUStrategy(min_used_freq_power)) {
+  if (lxu_strategy_type != "mixed_lru_lfu") {
```
divchenko: same as above, this check is odd here.

torchrec/csrc/dynamic_embedding/details/id_transformer_variant.h
```diff
@@ -0,0 +1,98 @@
...
+
+ private:
+  Variant strategy_;
+};
+
+class IDTransformerVariant {
+  // Currently, only NaiveIDTransformer is the IDTransformerVariant. Add more
+  // strategy when it is necessary.
+  using Variant = std::variant<NaiveIDTransformer<uint32_t>>;
+
```
divchenko: I think it would be simpler not to use std::variant, but just to rely on virtual functions instead. This way you would just have IDTransformer and NaiveIDTransformer, which extends IDTransformer. It also would have been much easier to reason about it.

torchrec/csrc/dynamic_embedding/details/id_transformer_variant.h
```diff
@@ -0,0 +1,98 @@
...
+#include <span>
+#include <type_traits>
+#include <variant>
+
+namespace torchrec {
+
+class LXUStrategyVariant {
+  // Currently, only MixedLFULRUStrategy is the LXUStrategyVariant. Add more
+  // strategy when it is necessary.
+  using Variant = std::variant<MixedLFULRUStrategy>;
```
divchenko: Same as below - it would be simpler to use class inheritance and virtual functions instead of std::variant.

divchenko (approved):
Suggested revising idtransformer/strategy class hierarchy (in follow-up PRs).

zhuzilin:
@divchenko I'm afraid we cannot change the variant into inheritance. The reason is that we are using template methods like:
```c++
template <typename Update, typename Fetch>
inline bool NaiveIDTransformer::transform(
    std::span<const int64_t> global_ids,
    std::span<int64_t> cache_ids,
    Update update,
    Fetch fetch)
```
so that the compiler could inline `update` and `fetch` callback to achieve better performance.

I wonder if we could change the code's structure without removing the variant?

divchenko:
LGTM, Thanks!

PR #897, opened 2022-12-15
bump version and pin requirements for release; add 3.10 to upload

colin2328:
Summary: as title

Differential Revision: D42057947


dzhulgakov:
@colin2328 - what was the rationale for pinning many packages to very recent versions? I'm running into some conflicts with other libraries. E.g. latest available packages for `nvtabular/cudf` don't seem to work with pyarrow=10, only with 9. However TorchRec pins the version. I bet some of these constraints are safe to relax.

Also the list is way too long itself. Many packages (e.g. aforementioned pyarrow or pondalocker) don't even get referenced in the source code.

colin2328:
Hi @dzhulgakov 
The rationale was that the versions were too broad (no versions), and we wanted to pin to ensure reliability. Part of this to make sure that DLRMv2 works reliably (the new ML perf is on TorchRec). 

Its likely true that this is too strict, if you want to relax, feel free to put out a PR and we can test.


I believe pyarrow is not used in source code, only in example, and yes we can probably remove and add to the README of the example.

PR #932, opened 2023-01-11
skip tests which are either hanging or not supported now on rocm

bmedishe:
skip tests that are hanging or currently not supported on rocm


colin2328 (commented):
Hi @bmedishe 
Thanks for the PR. Is this working for you?

bmedishe:
> Hi @bmedishe Thanks for the PR. Is this working for you?

@colin2328  Yes , these are  tests which are currently failing on rocm, we are working on enabling these tests too, till then they are skipped (for torchrec to work on rocm , fbgemm_gpu has to be installed from source , not pip install)

colin2328:
@bmedishe makes sense, I saw this from fbgemm_gpu team (support only for source install).
Thanks for working on this. Once this works, and you have tested , we can add instructions to torchrec README for experimental support for ROCm (i.e supported by fbgemm from source, link to their install )

LMK when PR is ready for reiew

PR #940, opened 2023-01-13
Remove the need to use populate_fx_names

divchenko:
Just call .forward() method directly. This way fx tracer won't fail checking that module is registered as a valid sub-module.


YLGH (approved):
Oh that's pretty slick, 

thanks!

PR #944, opened 2023-01-14
Mitigate feature processor hack

divchenko:
It was meant to support DMP, but named_modules should work on DMP just fine. On another hand, if I pass a model w/o DMP being a top-level model, it will break.

### DIFF
--- a/torchrec/distributed/train_pipeline.py
+++ b/torchrec/distributed/train_pipeline.py
@@ -396,10 +396,16 @@ def _rewrite_model(  # noqa C901
     context: TrainPipelineContext,
     dist_stream: Optional[torch.cuda.streams.Stream],
 ) -> List[ShardedModule]:
-
     # Get underlying nn.Module
     if isinstance(model, DistributedModelParallel):
         model = model.module
+
+    # Collect feature processors.
+    for _, m in model.named_modules():
+        if isinstance(m, BaseGroupedFeatureProcessor):
+            context.feature_processor_forwards.append(m.forward)
+            # pyre-ignore[8]: Incompatible attribute type
+            m.forward = lambda x: x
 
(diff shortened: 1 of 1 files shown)


torchrec/distributed/train_pipeline.py
```diff
@@ -498,7 +498,7 @@ def _replace_fp_forward(self, model: torch.nn.Module) -> None:
                 m.forward = lambda x: x
 
     def _connect(self, dataloader_iter: Iterator[In]) -> None:
-        self._replace_fp_forward(cast(torch.nn.Module, self._model.module))
+        self._replace_fp_forward(cast(torch.nn.Module, self._model))
```
YLGH: can we move logic from L401 to init instead? 

also, what does GMP stand for?

divchenko: good point, the hack is already there (to check for DMP). Would be nice to remove it altogether, but it might break in some cases...

divchenko: I meant DMP :)

xing-liu:
Should we just remove all feature preprocessing logics since nobody is using it?

divchenko:
> Should we just remove all feature preprocessing logics since nobody is using it?

Oh really? That would be great!!! Let's do in the follow-up large PR.

YLGH (approved):

YLGH:
naive thought, hadn't looked into feature p impl too much, but, can we separate out FP graph rewrite from train_pipeline entirely? (without deleting it?)

So then something like
```
model = SparseNNWithFeatureProcessor(...)
model = optimize_feature_processor(model)
for batch in SparseDistDataPipe(model, data_iter):
   ...
```

PR #945, opened 2023-01-14
Make TrainPipeline iterable/iterator

divchenko:
Basically this makes the following training loop possible:
```
pipeline = TrainPipeline(data_iter)
for batch in pipeline:
  res = model(batch)
  ...
```

Currently it's:
```
pipeline = TrainPipeline()
while True:
  try:
    res =model(pipeline.progress(data_iter))
  except StopIteration:
    break
```

__iter__ call doesn't return a reset iterator, but it's K. It's consistent w/ other Python APIs, e.g. itertools.count() does same thing.

Will add unit tests, if change looks good.

### DIFF
--- a/torchrec/distributed/tests/test_train_pipeline.py
+++ b/torchrec/distributed/tests/test_train_pipeline.py
@@ -12,6 +12,7 @@ from typing import cast, Dict, List, Optional, Tuple
 
 import torch
 import torch.distributed as dist
+from hypothesis import given, settings, strategies as st, Verbosity
 from torch import nn, optim
 from torchrec.distributed import DistributedModelParallel
 from torchrec.distributed.embedding_types import EmbeddingComputeKernel, KJTList
@@ -119,12 +120,14 @@ class TrainPipelineBaseTest(unittest.TestCase):
         torch.backends.cudnn.allow_tf32 = False
         torch.backends.cuda.matmul.allow_tf32 = False
 
-    # pyre-fixme[56]: Pyre was not able to infer the type of argument
     @unittest.skipIf(
         not torch.cuda.is_available(),
         "Not enough GPUs, this test requires at least one GPUs",
     )
-    def test_equal_to_non_pipelined(self) -> None:
+    # pyre-fixme[56]
+    @given(iter_api=st.sampled_from([True, False]))
+    @settings(verbosity=Verbosity.verbose, deadline=None)
+    def test_equal_to_non_pipelined(self, iter_api: bool) -> None:
         model_cpu = TestModule()
         model_gpu = TestModule().to(self.device)
         model_gpu.load_state_dict(model_cpu.state_dict())
@@ -138,7 +141,12 @@ class TrainPipelineBaseTest(unittest.TestCase):
             for b in range(5)
         ]
         dataloader = iter(data)
-        pipeline = TrainPipelineBase(model_gpu, optimizer_gpu, self.device)
+        if iter_api:
+            pipeline = iter(
+                TrainPipelineBase(model_gpu, optimizer_gpu, self.device, dataloader)
+            )
+        else:
+            pipeline = TrainPipelineBase(model_gpu, optimizer_gpu, self.device)
 
         for example in data[:-1]:
             optimizer_cpu.zero_grad()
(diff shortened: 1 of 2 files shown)


YLGH:
I really like this, I made some changes on top this change to try to further modularize train_pipeline a bit more

(really messy, just a prototype)
https://github.com/pytorch/torchrec/pull/963

I'm trying to make it so that pipeline doesn't actually take control of any of the training loop, e.g. calling anything optimizer related, backwards, you can define your own loss_fn like canonical training loop

curious of your thoughts @divchenko

divchenko:
Updated test

xing-liu:
Will it affect the performance? I like the interface.

xing-liu:
> I really like this, I made some changes on top this change to try to further modularize train_pipeline a bit more
> 
> (really messy, just a prototype) #963
> 
> I'm trying to make it so that pipeline doesn't actually take control of any of the training loop, e.g. calling anything optimizer related, backwards, you can define your own loss_fn like canonical training loop
> 
> curious of your thoughts @divchenko

Would you like to sync up offline for the new train pipeline design? @divchenko @YLGH

divchenko:
> > I really like this, I made some changes on top this change to try to further modularize train_pipeline a bit more
> > (really messy, just a prototype) #963
> > I'm trying to make it so that pipeline doesn't actually take control of any of the training loop, e.g. calling anything optimizer related, backwards, you can define your own loss_fn like canonical training loop
> > curious of your thoughts @divchenko
> 
> Would you like to sync up offline for the new train pipeline design? @divchenko @YLGH

Current PR is just a syntax sugar for existing impl. For @YLGH 's proposal - let's talk offline, yes

torchrec/distributed/tests/test_train_pipeline.py
```diff
@@ -138,15 +138,17 @@ def test_equal_to_non_pipelined(self) -> None:
             for b in range(5)
         ]
         dataloader = iter(data)
-        pipeline = TrainPipelineBase(model_gpu, optimizer_gpu, self.device)
+        pipeline = iter(
```
YLGH: let's add a parallel test for both the new interface and maintain old one (have a lot of internal code to migrate :))

I ran into some tricky StopIteration issues when doing
for batch in pipeline:
   ....

instead of next(pipeline),

but maybe i did something weird

YLGH (changes requested):

torchrec/distributed/tests/test_train_pipeline.py
divchenko: I'll make parallel unit test.
you don't need to migrate existing code - keep using progress() for a while.

divchenko:
API is backwards-compatible, we can keep current progress() method for a while.

torchrec/distributed/train_pipeline.py
```diff
@@ -81,10 +82,12 @@ def __init__(
         model: torch.nn.Module,
         optimizer: torch.optim.Optimizer,
         device: torch.device,
+        dataloader_iter: Iterator[In] = None,
```
YLGH: Optional[Iterator[In]]

A couple of lint errors - I think 
```
pip install pre-commit
pre-commit
```

can catch them 

@divchenko

YLGH (approved):
LGTM - just doing a quick internal e2e test to validate 

@dstaay-fb for a second look

torchrec/distributed/train_pipeline.py
```diff
@@ -68,7 +69,7 @@ def _wait_for_batch(batch: In, stream: Optional[torch.cuda.streams.Stream]) -> N
     batch.record_stream(cur_stream)
 
 
-class TrainPipelineBase(TrainPipeline[In, Out]):
+class TrainPipelineBase(TrainPipeline[In, Out], Iterable[Out]):
     """
     This class runs training iterations using a pipeline of two stages, each as a CUDA
     stream, namely, the current (default) stream and `self._memcpy_stream`. For each
```
colin2328, line 75: please add args to the docstrings.

let's also clarify in the docstring (via example) that this can be used as iterator API via next (with passed in data-laoder_iter) , or via progress().

colin2328 (commented):
overall I like the changes, thanks @divchenko

torchrec/distributed/train_pipeline.py
divchenko: There is no really good docs right now... We are actually discussing a new fully componentized pipeline API, which will allow me to fully keep my original training loop. Let's document it very well as a part of this new API.

```
torchrec.dist.pipeline_data(data_pipe: IterDataPipe, model: nn.Module) -> Tuple[IterDataPipe, nn.Module]


data_pipe, model = torchrec.dist.pipeline_data(input_loader, model)

for batch in data_pipe:
  opt.zero_grad()
  res = model(batch)
  res.backward()
  opt.step()
```

divchenko:
Will be replaced w/ the new DataPipe-based API.

PR #946, opened 2023-01-14
Sharding plan pretty printing

divchenko:
Example output: https://gist.github.com/divchenko/6e9526ed7829d2fd5cf7afb2384509c2

We separate parameter sharding and shards tables as it's possible to have several shards per parameter.


YLGH (approved):

YLGH:
let's land directly to main instead of pipeline iterable?

YLGH:
one thing that's been bugging me is that we call parameters 
user_id, instead of user_id.weight, 

but it's pretty annoying to change at this point i guess

PR #947, opened 2023-01-15
Simplify shard() call

divchenko:
Simplify shard() call by supporting ParameterShardingGenerator. New code
```
ebc = EmbeddingBagCollection()
sharded_ebc = shard(ebc, table_row_wise(host_index=0))
```
vs current one:
```
ebc = EmbeddingBagCollection()
plan=construct_module_sharding_plan(
  ebc,
  apply_to_all(ebc, table_row_wise(host_index=0)),
)
sharded_ebc = shard(ebc, plan)
```

Will update test if looks good.

### DIFF
--- a/torchrec/distributed/composable/tests/test_ddp.py
+++ b/torchrec/distributed/composable/tests/test_ddp.py
@@ -24,11 +24,7 @@ from torch.distributed.checkpoint import (
 )
 from torch.distributed.launcher.api import elastic_launch, LaunchConfig
 from torchrec.distributed.shard import shard as trec_shard
-from torchrec.distributed.sharding_plan import (
-    apply_to_all,
-    column_wise,
-    construct_module_sharding_plan,
-)
+from torchrec.distributed.sharding_plan import column_wise
 from torchrec.distributed.test_utils.test_model import ModelInput, TestSparseNN
 from torchrec.modules.embedding_configs import EmbeddingBagConfig
 from torchrec.test_utils import skip_if_asan
@@ -76,22 +72,12 @@ class DDPTest(unittest.TestCase):
         m.sparse.ebc = trec_shard(
             module=m.sparse.ebc,
             device=device,
-            plan=construct_module_sharding_plan(
-                m.sparse.ebc,
-                apply_to_all(m.sparse.ebc, column_wise(ranks=list(range(world_size)))),
-                device_type=device.type,
-            ),
+            plan=column_wise(ranks=list(range(world_size))),
         )
         m.sparse.weighted_ebc = trec_shard(
             module=m.sparse.weighted_ebc,
             device=device,
-            plan=construct_module_sharding_plan(
-                m.sparse.weighted_ebc,
-                apply_to_all(
-                    m.sparse.weighted_ebc, column_wise(ranks=list(range(world_size)))
-                ),
-                device_type=device.type,
-            ),
+            plan=column_wise(ranks=list(range(world_size))),
         )
         m.over = replicate(m.over)
         m.dense = replicate(m.dense)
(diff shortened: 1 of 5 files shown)


YLGH:
oh nice, so this moves construct_module_sharding_plan + apply_to_all logic into shard, but can still use construct_module_sharding_plan to inspect plan

most common use case should still be Dict[str, ParameterShardingGenerator] right?

divchenko:
> oh nice, so this moves construct_module_sharding_plan + apply_to_all logic into shard, but can still use construct_module_sharding_plan to inspect plan
> 
> most common use case should still be Dict[str, ParameterShardingGenerator] right?

I would actually say that table-wise sharding by default is the most common case. Typically models fit on a single GPU, but we want to speed them up by running training on multiple GPUs. Embedding tables are not tiny, so table-wise sharding will win over default data parallel (full copy of emb parameters).

YLGH:
> > oh nice, so this moves construct_module_sharding_plan + apply_to_all logic into shard, but can still use construct_module_sharding_plan to inspect plan
> > most common use case should still be Dict[str, ParameterShardingGenerator] right?
> 
> I would actually say that table-wise sharding by default is the most common case. Typically models fit on a single GPU, but we want to speed them up by running training on multiple GPUs. Embedding tables are not tiny, so table-wise sharding will win over default data parallel (full copy of emb parameters).

I mean for the input to shard(

you probably still want something (mostly) like

sharded_ebc = shard(ebc, {
 "table_0" : table_wise(0),
 "table_1":  table_wise(1),
})

instead of 

sharded_ebc = shard(ebc, table_wise(0))

right?

xing-liu:
shard(...) can only shard embedding modules? Why don't we make it a member function of EBC?

divchenko:
> shard(...) can only shard embedding modules? Why don't we make it a member function of EBC?

I guess we don't want to have non-sharded module code depend on torchrec.distributed...

divchenko:
> table_wise

I think both cases should be supported.

xing-liu:
> > shard(...) can only shard embedding modules? Why don't we make it a member function of EBC?
> 
> I guess we don't want to have non-sharded module code depend on torchrec.distributed...

Makes sense. We're trying to add new shardable and sharded modules (non-embedding) for retrievial models. The two problems are shard(...) can only shard embedding modules and non-embedding modules' shard(...) will have different signature (eg, don't need sharding plan). Do you have any ideas for a better design w/o make shard(...) a member function?

Related, the current planner is actually embedding planner which only works for embedding modules.

YLGH:
> > > shard(...) can only shard embedding modules? Why don't we make it a member function of EBC?
> > 
> > 
> > I guess we don't want to have non-sharded module code depend on torchrec.distributed...
> 
> Makes sense. We're trying to add new shardable and sharded modules (non-embedding) for retrievial models. The two problems are shard(...) can only shard embedding modules and non-embedding modules' shard(...) will have different signature (eg, don't need sharding plan). Do you have any ideas for a better design w/o make shard(...) a member function?
> 
> Related, the current planner is actually embedding planner which only works for embedding modules.

maybe we just call current shard, shard_embeddings, and then for non_parameterized things that don't need ParameterSharding, call them shard/shard_tensors?

divchenko:
> > > > shard(...) can only shard embedding modules? Why don't we make it a member function of EBC?
> > > 
> > > 
> > > I guess we don't want to have non-sharded module code depend on torchrec.distributed...
> > 
> > 
> > Makes sense. We're trying to add new shardable and sharded modules (non-embedding) for retrievial models. The two problems are shard(...) can only shard embedding modules and non-embedding modules' shard(...) will have different signature (eg, don't need sharding plan). Do you have any ideas for a better design w/o make shard(...) a member function?
> > Related, the current planner is actually embedding planner which only works for embedding modules.
> 
> maybe we just call current shard, shard_embeddings, and then for non_parameterized things that don't need ParameterSharding, call them shard/shard_tensors?

Hm, shard() call seems quite generic. Maybe make the plan optional if it's fixed. Curious to see design for retrieval model sharding @xing-liu

YLGH (approved):

colin2328:
Thanks @divchenko ! looks good

@xing-liu  -> shard() is not tied to embeddings, but rather tied to sharding ShardableModules(). As long as the retrieval modules are ShardableModules() and have an associated sharder, there shouldn't be any incompatibility.
LMK if something is off here

PR #955, opened 2023-01-18
Use the string value in `pooling_type_to_pooling_mode()`

842974287:
Summary:
If the torchrec model is from torch package, `pooling_type_to_pooling_mode()` won't work unless it's coming from the same torch package as well.

Changing to compare the string value works.

Differential Revision: D42585412


houseroad (approved):
LGTM

PR #963, opened 2023-01-20
Generic train pipeline

YLGH:
Summary:
factored so that train_pipeline is only responsible for the overlap, user will define anything else, this lets you do any optimizer.steps(), optimizer.forward, and you don't have to restrict model output to be (losses, output) anymore you can make it just output and do the loss separately

This lets you have the following training code:

```

        pipeline = TrainPipelineSparseDist(distributed_model, self.device, dataloader)

        for forward_ctx_gen in pipeline:
            with forward_ctx_gen() as forward_ctx:
                with record_function("## zero_grad ##"):
                    optimizer_distributed.zero_grad()

                def forward() -> None:
                    # TODO, can be called forward_ctx.batch
                    return distributed_model(forward_ctx._batch_i)

                (_losses, preds) = forward_ctx.run(forward)
                losses = torch.nn.functional.binary_cross_entropy_with_logits(
                    preds, forward_ctx._batch_i.label
                )
                print("preds", preds)
                print("losses", losses)

                all_loss.append(losses)
                with record_function("## backward ##"):
                    torch.sum(losses, dim=0).backward()
                with record_function("## optimizer ##"):
                    optimizer_distributed.step()
```

Differential Revision: D42623963

### DIFF
--- a/torchrec/distributed/tests/test_train_pipeline.py
+++ b/torchrec/distributed/tests/test_train_pipeline.py
@@ -13,6 +13,7 @@ from typing import cast, Dict, List, Optional, Tuple
 import torch
 import torch.distributed as dist
 from torch import nn, optim
+from torch.autograd.profiler import record_function
 from torchrec.distributed import DistributedModelParallel
 from torchrec.distributed.embedding_types import EmbeddingComputeKernel, KJTList
 from torchrec.distributed.embeddingbag import (
@@ -97,6 +98,9 @@ class ModelInputSimple(Pipelineable):
         self.float_features.record_stream(stream)
         # pyre-fixme[6]: For 1st param expected `Stream` but got `Stream`.
         self.label.record_stream(stream)
+
+
+import logging
 
 
 class TestModule(nn.Module):
@@ -198,9 +202,6 @@ class TrainPipelineSparseDistTest(unittest.TestCase):
             dict(in_backward_optimizer_filter(distributed_model.named_parameters())),
             lambda params: optim.SGD(params, lr=0.1),
         )
-        pipeline = TrainPipelineSparseDist(
-            distributed_model, optimizer_distributed, self.device
-        )
 
         data = [
             ModelInput.generate(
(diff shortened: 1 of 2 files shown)


divchenko:
Great observation! We currently have to re-write TrainPipeline when we compute multiple losses, etc.
We indeed zero grad before any other ops in TrainPipeline and backward/opt.step after all ops.
I actually would keep it a bit simpler. Just do the following in TrainPipeline:
```
...
ret =  self._model(batch_i)
# Data Distribution
return ret
```

then the code becomes (we probably should call it a different name - DataPipeline as it no longer does any training):

```
pipeline = DataPipeline(model, data_iter)
opt = torch.optim.SGD(...)
for batch in pipeline:
  opt.zero_grad()
  loss, output = pipeline(batch)
  loss.backwad()
  opt.step()
  ...
```

Let's create a separate class. We can keep current TrainPipeline as is (it's a special case, but it's very common and saves some boilerplate training code - think Lightninig :) ). Then we make TrainPipeline use DataPipeline.

YLGH:
> Great observation! We currently have to re-write TrainPipeline when we compute multiple losses, etc. We indeed zero grad before any other ops in TrainPipeline and backward/opt.step after all ops. I actually would keep it a bit simpler. Just do the following in TrainPipeline:
> 
> ```
> ...
> ret =  self._model(batch_i)
> # Data Distribution
> return ret
> ```
> 
> then the code becomes (we probably should call it a different name - DataPipeline as it no longer does any training):
> 
> ```
> pipeline = DataPipeline(model, data_iter)
> opt = torch.optim.SGD(...)
> for batch in pipeline:
>   opt.zero_grad()
>   loss, output = pipeline(batch)
>   loss.backwad()
>   opt.step()
>   ...
> ```
> 
> Let's create a separate class. We can keep current TrainPipeline as is (it's a special case, but it's very common and saves some boilerplate training code - think Lightninig :) ). Then we make TrainPipeline use DataPipeline.

oh yeah - a lot of the ugliness comes from the custom model(batch) call, but I doubt this will be a common use case :)

I like the idea of splitting this out into separate class first

PR #970, opened 2023-01-23
Minimize dependencies

divchenko:
1. Split dependencies into runtime (install-requirements.txt) and dev time dependencies (requirements.txt). 
2. Add ability to override version during build.

These dependencies I just can't find anywhere to be used:
```
arrow
attrs
certifi
charset-normalizer
Cython
docker
docstring-parser
filelock
fsspec
idna
Jinja2
markupsafe - only packaging mock in inference packager
mypy-extensions
ninja
packaging
portalocker
pyDeprecate
pyparsing
python-dateutil
pytz
PyYAML
requests
six - only packaging mock in inference packager
sortedcontainers
typing-inspect
typing_extensions
urllib3
websocket-client
```

These are runtime dependencies:
```
fbgemm-gpu-nightly
iopath - only needed by torchrec.datasets
pandas - only needed by torchrec.datasets and examples
tabulate
torchmetrics
tqdm  - only needed by torchrec.datasets
```

These dependencies are needed for dev and runtime:
```
black - missing
cmake
fbgemm-gpu-nightly
hypothesis
iopath
numpy
pandas
pyre-extensions
scikit-build
tabulate
torcharrow
torchmetrics
torchx - only in examples
tqdm - only needed by torchrec.datasets and examples
```


YLGH:
I think we just copied requirements from another library :/ , thanks for the clean up

I recall @colin2328  pinned some of the versions for DLRMv2 - @colin2328  is this still necessary? 

I think pyre-extensions had a backward breaking change (maybe some >=3.8 feature) so let's keep that pinned 
(https://github.com/pytorch/torchrec/pull/433)

colin2328:
Thanks @divchenko . we will test these out.

I think the original requirements were pip frozen (when we built fbgemm from source) from intitial setup and haven't been updated much.

@YLGH  for DLRMv2, yes we did pin based on their request. They are pinned now to 0.3.2 so no new code should affect them.

Talked to @osalpekar , I think we should follow what torchvision other libs do which is to do what @divchenko  is doing here (split into reqs.txt and instal-reqs.txt) and relax as. much as possible. We will test these out, but need to fix the CI first

setup.py
```diff
@@ -66,7 +72,9 @@ def main(argv: List[str]) -> None:
         reqs = f.read()
         install_requires = reqs.strip().split("\n")
```
osalpekar, line 73: I think we'll need to parse `install-requirements.txt` above instead of `requirements.txt` to have this change propagate to the `setup.py` wheels build

osalpekar (commented):
Thanks! @colin2328 @YLGH Perhaps in addition to manual testing, we should have the wheels build GHA run per-PR (without uploading) so we can easily test out such changes. Other domains libs do this as well

setup.py
divchenko: oh, forgot to update, good catch!

requirements.txt
```diff
@@ -1,43 +1,14 @@
...
-typing-inspect>=0.8.0
-typing_extensions>=4.4.0
-urllib3>=1.26.13
-usort>=1.0.5
-websocket-client>=1.4.2
+hypothesis
+iopath
+numpy
+pandas
+pyre-extensions
```
colin2328: pyre-extensions==0.0.27

colin2328 (approved):
thanks!

osalpekar (approved):
LGTM

PR #987, opened 2023-01-27
Enable data pipelining

lequytra:
Summary:
Note: Work in progress, publish for early feedbacks.

Currently, torchrec's built-in `TrainPipeline` API takes full control of the training loop (i.e. load next batch, forward, backward, optimizer step).

While this satisfies most common use cases, researchers who want to experiment with more advanced techniques (e.g. multi-losses, multiple batches) will have to write their own custom training loop, often with lots of duplicate codes to take advantage of comm & comp overlap speedup.

Goal of `DataPipeline` is to provide easy API to enable comm & comp overlap, but gives users back control over the training loop.

Differential Revision: D42810112

### DIFF
--- a/torchrec/distributed/data_pipeline.py
+++ b/torchrec/distributed/data_pipeline.py
@@ -5,12 +5,13 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import abc
+
 import logging
-from typing import cast, Dict, Generic, Iterator, List, Optional, Tuple, TypeVar
+from typing import cast, Dict, Iterator, List, Optional, TypeVar
 
 import torch
-from torch.autograd.profiler import record_function
+
+from torch.profiler import record_function
 from torchrec.distributed.model_parallel import ShardedModule
 from torchrec.distributed.pipeline_utils import (
     _rewrite_model,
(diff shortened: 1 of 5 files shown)


divchenko:
Hey, 
@lequytra 
 I actually was thinking about something like: https://github.com/pytorch/torchrec/blob/data_pipeline/torchrec/distributed/data_pipeline.py
This way way training loop can totally stay intact - users will call model's forward methods exactly same way as before.
I've not tested these changes, I thought it was just easier to write it vs trying to explain in words :)

torchrec/distributed/data_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+# pyre-ignore
+def module_overlap_forward_hook(module: torch.nn.Module, args, kwargs) -> None:
+    assert hasattr(
+        module, "_datapipeline"
+    ), "No datapipeline attached to the module, cannot perform data pipelining."
+    # pyre-ignore
+    module._datapipeline._module_overlap_forward_hook()
+
+
+class DataPipeline(Iterator[In]):
```
divchenko, line 30: why do we need this base class?  It's basically Iterator

torchrec/distributed/data_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+from torchrec.distributed.train_pipeline import _start_data_dist
+from torchrec.streamable import Pipelineable
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+In = TypeVar("In", bound=Pipelineable)
+
+
+# pyre-ignore
+def module_overlap_forward_hook(module: torch.nn.Module, args, kwargs) -> None:
```
divchenko: prefix w/ _ ?

divchenko (approved):
LGTM

vargab100:
ModuleNotFoundError: No module named 'data_pipeline
getting this error for .....
from  data_pipeline import transformation_pipeline
this line 
i try !pip install data_pipeline   command 
but no improvement 
anyone pls say is data_pipeline is removed? then what alternative i have to do

torchrec/distributed/data_pipeline.py
lequytra: Between inheriting `Generic[In]` and `Iterator[In]`, I used `Iterator[In]` for readability and enforce that `DataPipeline` stays an iterator.

torchrec/distributed/data_pipeline.py
lequytra: hook is removed in updated version.

lequytra:
> 



> ModuleNotFoundError: No module named 'data_pipeline getting this error for ..... from data_pipeline import transformation_pipeline this line i try !pip install data_pipeline command but no improvement anyone pls say is data_pipeline is removed? then what alternative i have to do

Hey @vargab100, our library doesn't have `transformation_pipeline`, and the new `data_pipeline` feature is still in PR and has not been merged yet. Maybe it's a different library, you can reach out to the library PoCs about this issue.

rllin:
waiting eagerly for this to pull in, is the intention still to finish and merge this?

PR #991, opened 2023-01-27
speed up TorchRec CPU unit tests by only installing CPU fbgemm; also use fail-fast strategy

colin2328:
Differential Revision: D42815383

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,12 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
         default=None,
         help="override version",
     )
+    parser.add_argument(
+        "--cpu-only",
+        type=bool,
+        default=False,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -82,7 +88,16 @@ def main(argv: List[str]) -> None:
...
         install_requires.append("fbgemm-gpu")
 
-    print(f"-- {name} building version: {version}")
+    cpu_only = args.cpu_only
+    if cpu_only:
+        if "fbgemm-gpu-nightly" in install_requires:
+            install_requires.remove("fbgemm-gpu-nightly")
+            install_requires.append("fbgemm-gpu-nightly-cpu")
+        if "fbgemm-gpu" in install_requires:
+            install_requires.remove("fbgemm-gpu")
```
colin2328, line 97: see here @divchenko

PR #992, opened 2023-01-28
pyre setup

divchenko:
Moved CopyableMixin to distributed.utils (where it should be as it relies on distributed.utils. copy_to_device call).
We should also add examples/contrib/benchmarks eventually (they are not even checked by FB internal typechecker)


colin2328 (commented):
@divchenko can you add a section to README for external development and how to run pyre

divchenko:
> @divchenko can you add a section to README for external development and how to run pyre

oh yeah, I'll do in the follow-up + will add github action

YLGH:
changes were merged in separate PR

PR #997, opened 2023-01-31
pyre datasets

divchenko:
- pyre for datasets
- added running 'pyre check' into pre-commit workflow


divchenko:
Merged into https://github.com/pytorch/torchrec/pull/992

PR #998, opened 2023-01-31
pyre hook

divchenko:
- pyre setup
- Pyre check GH action


divchenko:
Merged into https://github.com/pytorch/torchrec/pull/992

PR #1003, opened 2023-02-01
Fixing pyre issues

YLGH:
Summary: refactor and address some pyre issues

Differential Revision: D42905705

### DIFF
--- a/.github/workflows/pyre.yml
+++ b/.github/workflows/pyre.yml
@@ -1,0 +1,27 @@
+name: Pyre Check
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  pyre-check:
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash -el {0}
+    steps:
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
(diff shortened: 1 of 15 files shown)


.github/workflows/pyre.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: 3.8
+      - name: Checkout Torchrec
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: >
+          conda install --yes -c pytorch-nightly "pytorch-nightly"::pytorch[build="*cpu*"] &&
+          pip install -r requirements.txt &&
+          pip uninstall fbgemm_gpu-nightly -y &&
```
colin2328: rather than do this, we can replace with CPU: sed -i 's/fbgemm-gpu-nightly/fbgemm-gpu-nightly-cpu/g' requirements.txt

YLGH: sure, but is there a benefit to doing this?

YLGH: ah.- I guess it saves an install

PR #1015, opened 2023-02-06
Remove flags from setup.py build

osalpekar:
Starting to make torchrec play better with Nova builds by removing flag args from setup.py build. The Nova workflows don't take any args to the `setup.py` script, so any feature toggling is done via env vars (which can be configured for each domain library).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,10 +38,8 @@ def get_channel():
     return os.getenv("CHANNEL")
 
 
-def get_CPU_only():
-    cpu_only = os.getenv("CPU_ONLY", "0")
-    assert cpu_only in ["0", "1"]
-    return int(cpu_only)
+def get_cu_version():
+    return os.getenv("CU_VERSION", "cpu")
 
 
 def parse_args(argv: List[str]) -> argparse.Namespace:
@@ -79,8 +77,8 @@ def main(argv: List[str]) -> None:
             install_requires.remove("fbgemm-gpu-nightly")
         install_requires.append("fbgemm-gpu")
 
-    cpu_only = get_CPU_only()
-    if cpu_only:
+    cu_version = get_cu_version()
+    if cu_version == "cpu":
         if "fbgemm-gpu-nightly" in install_requires:
             install_requires.remove("fbgemm-gpu-nightly")
             install_requires.append("fbgemm-gpu-nightly-cpu")
@@ -88,7 +86,7 @@ def main(argv: List[str]) -> None:
             install_requires.remove("fbgemm-gpu")
             install_requires.append("fbgemm-gpu-cpu")
 
-    print(f"-- {name} building version: {version} CPU only: {bool(cpu_only)}")
+    print(f"-- {name} building version: {version} CU Version: {cu_version}")
 
     packages = find_packages(
         exclude=(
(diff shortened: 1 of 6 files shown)


colin2328:
cc @divchenko FYI

osalpekar:
Quick Update on this PR - I'm avoiding making too many changes close to the 2.0 RC cut. Once the 2.0 release is closer to being done, I'll come back to the build improvements here :)

osalpekar:
This does fix the CPU wheels build: https://github.com/pytorch/torchrec/actions/runs/5871439740/job/15920795895?pr=1015. Going to comment out the line that triggers the Nova wheels builds on PR's before landing.

henrylhtsang:
> This does fix the CPU wheels build: https://github.com/pytorch/torchrec/actions/runs/5871439740/job/15920795895?pr=1015. Going to comment out the line that triggers the Nova wheels builds on PR's before landing.

BTW I think the cuda 11.8, which was briefly fixed, is failing again due to fbgemm https://github.com/pytorch/FBGEMM/issues/1847

henrylhtsang:
Can we fix this as well? I know it was caused by me

https://github.com/pytorch/torchrec/blob/main/.github/workflows/docs.yml#L58-L60

edit: I guess also pyre.yml to some degree, but we can handle them in another diff

setup.py
```diff
@@ -79,7 +68,7 @@ def main(argv: List[str]) -> None:
         reqs = f.read()
         install_requires = reqs.strip().split("\n")
 
-    version = args.version
+    version = os.getenv("BUILD_VERSION")
```
henrylhtsang, line 71: Our version on main is not updated to 0.4.0 I think. Is that going to be a problem?

https://github.com/pytorch/torchrec/blob/main/version.py

henrylhtsang (approved):
LGTM overall. A few nits.

osalpekar:
> Can we fix this as well? I know it was caused by me
> 
> https://github.com/pytorch/torchrec/blob/main/.github/workflows/docs.yml#L58-L60
> 
> edit: I guess also pyre.yml to some degree, but we can handle them in another diff

We can fix this once we enable the Nova pipeline. At that point we should just be able to replace those lines with : 
`pip install torchrec --index-url https://download.pytorch.org/whl/nightly/cpu`

osalpekar:
> > This does fix the CPU wheels build: https://github.com/pytorch/torchrec/actions/runs/5871439740/job/15920795895?pr=1015. Going to comment out the line that triggers the Nova wheels builds on PR's before landing.
> 
> BTW I think the cuda 11.8, which was briefly fixed, is failing again due to fbgemm [pytorch/FBGEMM#1847](https://github.com/pytorch/FBGEMM/issues/1847)

Strange, I got 11.8 working recently in https://github.com/pytorch/torchrec/pull/1329. Looks like it's also failing on the old pipelines (https://github.com/pytorch/torchrec/actions/runs/5878090975/job/15939592873) so it's not Nova-specific. Looks like a legit issue in FBGEMM?

setup.py
osalpekar: Feel free to updated it to the desired version. I've noticed in the past that `torchrec.__version__` hasn't worked correctly, so we may want to mimic what other domain repos like vision do for versioning: https://github.com/pytorch/vision/blob/main/setup.py.

PR #1067, opened 2023-03-05
Enrich the EmbeddingBagCollection class description by adding a more inclusive example

ldniubuzhi:
**What**
Change the class doc str to illustrate a more inclusive example.

- Changed 2nd embedding bag config to have multiple features that share the same underlying embeddings
- The embedding sharing is tested between the corresponding two features (i.e. f2 and f3)

**Note**
No change to the implementation code.


joshuadeng:
hi @ldniubuzhi can you address the pre-commit formatting error?

PR #1536, opened 2023-11-21
StreamSyncTensor implicit pipelining for TorchRec [Prototype]

YLGH:
Provide implicit pipelining in TorchRec by using streams. StreamSyncTensor will continue on its current stream if all the incoming tensors are StreamSync tensors on that Stream.

Otherwise it will sync to the main stream and continue.

Random thought/TODO - maybe all torch.CudaTensor should have this kind of semantic (to avoid any record_stream semantics)

### DIFF
--- a/torchrec/distributed/stream_sync_tensor.py
+++ b/torchrec/distributed/stream_sync_tensor.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from functools import partial
+from typing import Any, Tuple
+
+import torch
+from torch.autograd import Function
+from torch.utils._python_dispatch import return_and_correct_aliasing
+from torch.utils._pytree import tree_leaves, tree_map_only
+
+
+class StreamSyncTensor(torch.Tensor):
+    """
+    A Tensor wrapper subclass that is used to manage tensors from different streams.
+    This is particularly useful to cleanly overlap collective call operations with compute ops.
+
+    There are two main modes of the dispatch function.
+
+    1. If all the incoming tensors are StreamSyncTensors on the same steream, then no stream synchronizations are requires, and we just perform the op on the same stream.
+    2. Otherwise, we assume that all the tensors are on different streams, and we synchronize them before performing the operation.
+
+    args::
+        elem: torch.Tensor - torch.Tensor that may have been created from a separate stream.
+        stream: torch.cuda.Stream - stream that the tensor was created on.
+    Example::
+
+        s = torch.cuda.Stream()
+        with torch.cuda.stream(s):
+            x = torch.randn(10, device="cuda")
+            y = 2*x
+        y_stream = StreamSyncTensor(y, stream=s)
+
+        z = torch.randn(10, device="cuda")
+
+        # here y_stream will manage the syncing of stream s and current cuda stream, as well as any record_stream semantics.
+        a = y_stream + z
+    """
+
+    __slots__ = ["elem", "stream"]
+    # pyre-ignore
+    __torch_function__ = torch._C._disabled_torch_function_impl
+
+    @staticmethod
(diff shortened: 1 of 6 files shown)


YLGH:
Copying over comments from diff from @ezyang 

```
I'm thinking about whether or not we can implement this class without use of record_stream (which can lead to suboptimal memory usage in the CUDA caching allocator.) The key ingredient is that we have to track uses of the tensor from other streams and ensure we don't deallocate our tensor before those uses are done. If we assume that the internal elem tensor of a StreamSyncTensor is private (no access that is not mediated through our subclass), it seems to me that we could have 100% accurate information on stream usage. We could then generate an event for each external stream usage representing events that must be complete before deallocation. We would then need to extend the CUDA caching allocator with an API that allows the user to add its own CUDA events instead of forcing a fresh event to be created at deallocation time:
  void insert_events(Block* block) {
    int prev_device = 0;
    C10_CUDA_CHECK(c10::cuda::GetDevice(&prev_device));

    stream_set streams(std::move(block->stream_uses));
    AT_ASSERT(block->stream_uses.empty());
    for (auto& stream : streams) {
      C10_CUDA_CHECK(c10::cuda::SetDevice(stream.device_index()));

      EventPool::Event event =
          create_event_internal(static_cast<int>(stream.device_index()));
      C10_CUDA_CHECK(cudaEventRecord(*event, stream.stream()));

      block->event_count++;
      cuda_events[stream].emplace_back(std::move(event), block);
    }

    C10_CUDA_CHECK(c10::cuda::MaybeSetDevice(prev_device));
  }
This would allow for more prompt deallocation, as the amount of time we have to wait would be solely until the known uses are complete, rather than whatever the state of the main stream was at the time of deallocation.
```

```
Ah, so this is a policy decision. I suppose it's not unreasonable to propagate these policy decisions via data-dependent values (e.g., similar to requires_grad), but I would ask if it makes more sense to have a context manager control the policy decision instead. For example, you could imagine:
with no_sync_to_main():
    x = all2all(...)
    y = post_all2all(x)
and this would result in recordStream calls on all the tensors post_all2all accessed that weren't stream tensors.
```

```
I do wonder if it is good UX to implicitly setup these waits. You could imagine that what if the SST is far ahead of the main stream, then this implicit wait results in a big bubble. But the user side code change could have been quite innocuous; just happening to add a dependency on some tensor result. I suppose this is what you were asking me about having lots of streams; there's no way to save you if you put everything on the same stream, but if you had a lot of streams you could potentially perform this wait on yet another stream, and let work that didn't depend on this particular operation continue to go ahead.
In any case, we should consider whether or not it would be better to force an explicit wait. Something like stream_sync_tensor.wait_on(plain_tensor) which makes it clear that we're waiting on plain tensor on the side stream in user code. I don't know how much this goops up user code in your example.
```

ezyang:
cc @ngimel

torchrec/distributed/stream_sync_tensor.py
```diff
@@ -0,0 +1,157 @@
...
+                out = tree_map_only(
+                    torch.Tensor,
+                    partial(
+                        wrap,
+                        stream=stream_sync_tensor_stream,
+                    ),
+                    out,
+                )
+                return return_and_correct_aliasing(func, args, kwargs, out)
+
```
YLGH, line 126: another mode is continuing on this stream

```
	
        elif not should_sync_to_main:
            # We have some ops on main stream (e.g. post A2A ops/e.g. CW Permute) that we want to run in side stream
            # we don't want these ops to block dense ops
            stream_sync_tensor_stream.wait_stream(torch.cuda.default_stream())
            for non_stream_sync_tensor in non_stream_sync_tensors:
                non_stream_sync_tensor.record_stream(stream_sync_tensor_stream)

            with torch.cuda.stream(stream_sync_tensor_stream):
                out = func(*unwrapped_args, **unwrapped_kwargs)
                out = tree_map_only(
                    torch.Tensor,
                    partial(
                        wrap,
                        stream=stream_sync_tensor_stream,
                        should_sync_to_main=should_sync_to_main,
                    ),
                    out,
                )
                return return_and_correct_aliasing(func, args, kwargs, out)
```

PR #1556, opened 2023-12-05
Add dtype option for get_block_sizes_runtime_device

tissue3:
Summary: As titled. Do not hardcode dtype in get_block_sizes_runtime_device

Differential Revision: D51857139


jingsh (approved):
lgtm

PR #1599, opened 2023-12-21
Support padding indices in embedding collection modules

murphymatt:
Summary: We currently do not support this functionality in embedding collection operations. However, there are emerging use cases where this will be more useful.

Differential Revision: D52356440


Ye-Tian-Zero:
Hi @murphymatt , may I ask what's the progress of this feature?

PR #1604, opened 2024-01-02
Add pybind for AOTIModelContainerRunnerCpu and AOTIModelContainerRunnerCuda

desertfire:
Summary: Now we can allocate an AOTIModelContainerRunner object instead of relying on torch.utils.cpp_extension.load_inline. Also renamed AOTInductorModelRunner to AOTIRunnerUtil in this PR.

Differential Revision: D52485173


desertfire:
Replace https://github.com/pytorch/pytorch/pull/116269

PR #1624, opened 2024-01-11
Add staged train pipeline to torchrec

joshuadeng:
Summary:
Copying from internal design doc:

prototype generic training pipeline that can support arbitrary number of stages.

The design is illustrated in the figure below. A pipeline is composed of K stages, each one depends on its precedence. Different stage may be executed in different or the same streams, with multiple batches concurrently executed in the same iteration. For example, in the image, batch[0] is the oldest batch that has passed H2D, preproc, and SDD stages, and will be running through comp in the current iteration. Batch[1], on the other hand, is the second oldest batch that will execute SDD in the current iteration. Similarly for other batches.

 {F1150156522}

with this, 4 batches will be handled together in the same iteration, while each of them is under a different stage. When an iteration is done, there will be a advance step to copy newer data (slots larger index) to older data (slots with smaller index) so that they will be handled by the next stage in the next iteration.
Preview image from Diff Tool

For SDD handling, we currently wrap on top of existing Trec utilities. This part could potentially be improved in the future.

Some additional things on top of this:

* Adding fill callback for start sdd
* Adding await sdd as a callback
* Modifying progress to be walrus'able

Differential Revision: D51182804

### DIFF
--- a/torchrec/distributed/train_pipeline/__init__.py
+++ b/torchrec/distributed/train_pipeline/__init__.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from torchrec.distributed.train_pipeline.train_pipeline import (  # noqa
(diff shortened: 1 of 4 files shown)


YLGH:
lgtm

joshuadeng:
come back @YLGH

PR #1656, opened 2024-01-24
AUC metric optionally move concat to compute step instead of update through flag

zainhuda:
Summary:
Previously unlanded D51399384 due to issues with some models. Introducing it again but through an experimental flag as this optimization can benefit many models. This should not conflict with any models running AUC metric. The feature flag is akin to fusing task computation in RecMetrics as that feature as well does not work across all families of models. 

The change in this diff is:
We don't need to concatenate the tensor on every update step, since it is an expensive operation (creates a new tensor and allocates new memory every call, as tensors are contiguous) we can call tensor.concat on the compute step instead. Which happens every compute_interval_step batches. This optimization should boost performance of models using AUC with no regression in metric quality.

Differential Revision: D53027097

### DIFF
--- a/torchrec/metrics/auc.py
+++ b/torchrec/metrics/auc.py
@@ -6,10 +6,11 @@
 # LICENSE file in the root directory of this source tree.
 
 from functools import partial
-from typing import Any, cast, Dict, List, Optional, Type
+from typing import Any, Callable, cast, Dict, List, Optional, Tuple, Type
 
 import torch
 import torch.distributed as dist
+from torchmetrics.utilities.distributed import gather_all_tensors
 from torchrec.metrics.metrics_config import RecComputeMode, RecTaskInfo
 from torchrec.metrics.metrics_namespace import MetricName, MetricNamespace, MetricPrefix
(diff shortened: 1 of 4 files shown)


Andrew12Liu:
Looks sick

torchrec/metrics/auc.py
```diff
@@ -241,25 +254,56 @@ def update(
...
+                    # Remove the first element from predictions, labels, and weights
+                    [lst.pop(0) for lst in [w_preds, w_labels, w_weights]]
+                else:
+                    # Update the first element of predictions, labels, and weights
+                    # Off by one potentially - keeping legacy behaviour
+                    for attribute in [w_preds, w_labels, w_weights]:
+                        attribute[0] = attribute[0][:, diff:]
+                    self._num_samples -= diff
+        else:
+            # Using `self.predictions =` will cause Pyre errors.
```
Andrew12Liu, line 285: nice comment

Andrew12Liu:
Looking forward to the performance boost!

Andrew12Liu (approved):

PR #1659, opened 2024-01-25
fix world size 1 for 2D lengths

jiayisuse:
Summary: when world_size=1 and lengths is created as 2d, sharded EBC will skip flattening kjt lengths

Differential Revision: D53068678


houseroad (approved):
Add a unittest?

PR #1866, opened 2024-04-10
Wrap DP tables with DDP after reset parameters

henrylhtsang:
Summary:
# Problem

Problem is once we wrap DP tables with DDP, the parameters are the same, but not synced. So if we reset the parameters in the table without using the same manual seed, it could cause the DP table parameters to be different. 

In other words, DP tables would be initialized differently.

# Fix

There are a few ways to fix it. This is the way we believe to be least invasive and follow the spirit of the api.

What we do:
1. Delay the DDP wrapping to reset_parameters
2. rewrap everytime we call reset_parameters

Differential Revision: D55227979

### DIFF
--- a/torchrec/distributed/embedding.py
+++ b/torchrec/distributed/embedding.py
@@ -17,7 +17,7 @@ from itertools import accumulate
 from typing import Any, cast, Dict, List, MutableMapping, Optional, Type, Union
 
 import torch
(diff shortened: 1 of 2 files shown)


henrylhtsang:
cc @YLGH

YLGH:
my bad

PR #1902, opened 2024-04-19
embedding vector init norm by embedding dim instead of num_embeddings

di-wnd:


di-wnd:
cc @PaulZhang12

TroyGarden (approved):
agree with the change: the init weight should be over emb_dim not emb_num

PR #1947, opened 2024-05-03
Fix dynamic embedding bugs that may lead to segmentation faults

sherlockkenan:
Hi TorchRec Team,

We've recently been trying to use the dynamic embedding feature in torchrec contrib, but we  have encountered a few challenges. The process may result in a segmentation fault. After debugging, we've identified two potential problems.

1） fetch_notifications_ is not thread-safe.
2）Tensor data might be recycled during pushing operations.

For the first issue, the fetch_notifications_ variable is accessed in both the pull and sync fetch functions. As these functions are called in different threads, there might be a thread safety issue.

For the second issue, the tensor generated by concat in the push function is a temporary variable. The subsequent call to io.push is asynchronous and does not copy the data. Therefore, the data might have been recycled by the time it is executed asynchronously, leading to access to invalid data.

In light of these findings, we have undertaken some corrective measures to address these issues. We appreciate your attention to this Pull Request and eagerly await your valuable feedback.

Thank you for your time and consideration.

### DIFF
--- a/contrib/dynamic_embedding/src/tde/ps.cpp
+++ b/contrib/dynamic_embedding/src/tde/ps.cpp
@@ -25,9 +25,12 @@ c10::intrusive_ptr<FetchHandle> PS::Fetch(
   if (cache_ids_to_fetch_or_evict_.empty()) {
     return c10::make_intrusive<FetchHandle>(time, c10::intrusive_ptr<PS>());
   }
-  fetch_notifications_.emplace_back(time, c10::make_intrusive<Notification>());
-  c10::intrusive_ptr<Notification> notification =
-      fetch_notifications_.back().second;
+  c10::intrusive_ptr<Notification> notification;
+  {
+    std::unique_lock<std::mutex> lock_fetch(fetch_notifications_mutex_);
+    fetch_notifications_.emplace_back(time, c10::make_intrusive<Notification>());
+    notification=fetch_notifications_.back().second;
+  } 
   uint32_t num_os_ids = os_ids_.size();
   io_.Pull(
       table_name_,
@@ -122,6 +125,9 @@ void PS::Evict(torch::Tensor ids_to_evict) {
     torch::Tensor data = torch::cat(all_tensors, 0).cpu();
     TORCH_CHECK(data.numel() == data_size * col_size_);
 
+    // to prevent the original data from being prematurely recycled
+    auto data_shared_ptr = std::make_shared<torch::Tensor>(data);
+
     offsets[0] = 0;
     for (uint32_t j = 0; j < all_tensors.size(); ++j) {
       offsets[j + 1] =
@@ -136,22 +142,30 @@ void PS::Evict(torch::Tensor ids_to_evict) {
         col_ids,
         os_ids_,
         tcb::span{
-            reinterpret_cast<uint8_t*>(data.data_ptr<float>()),
+            reinterpret_cast<uint8_t*>(data_shared_ptr->data_ptr<float>()),
             data_size * sizeof(float)},
         tcb::span{offsets.data(), offsets_size},
-        [&notification] { notification.Done(); });
+        [&notification, data_shared_ptr] { notification.Done(); });
   }
   notification.Wait();
 }
 
 void PS::SyncFetch(int64_t time) {
-  while (!fetch_notifications_.empty()) {
-    auto& [t, notification] = fetch_notifications_.front();
-    if (t != time && time >= 0) {
+  std::unique_lock<std::mutex> lock(
+      fetch_notifications_mutex_, std::defer_lock);
+
+  while (true) {
+    lock.lock();
+    if (fetch_notifications_.empty() ||
+        fetch_notifications_.front().first != time && time >= 0) {
+      lock.unlock();
       break;
     }
-    notification->Wait();
+    auto notification = fetch_notifications_.front().second;
     fetch_notifications_.pop_front();
+    lock.unlock();
+
+    notification->Wait();
   }
 }
 
--- a/contrib/dynamic_embedding/src/tde/ps.h
+++ b/contrib/dynamic_embedding/src/tde/ps.h
@@ -99,6 +99,7 @@ class PS : public torch::CustomClassHolder {
   void Filter(const torch::Tensor& tensor);
 
   std::mutex mu_;
+  std::mutex fetch_notifications_mutex_;
   std::string table_name_;
   c10::intrusive_ptr<LocalShardList> shards_;
   int64_t col_size_;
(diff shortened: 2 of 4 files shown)


henrylhtsang:
Hi, I don't have expertise on dynamic embeddings, so probably will just cc @colin2328 

also probably make an issue as well, since it is easier to track there

sherlockkenan:
sorry for the late response. I have created an issue here: https://github.com/pytorch/torchrec/issues/2006. I would appreciate any further suggestions regarding the pull request.

contrib/dynamic_embedding/src/tde/ps.cpp
```diff
@@ -25,9 +25,12 @@ c10::intrusive_ptr<FetchHandle> PS::Fetch(
   if (cache_ids_to_fetch_or_evict_.empty()) {
     return c10::make_intrusive<FetchHandle>(time, c10::intrusive_ptr<PS>());
   }
-  fetch_notifications_.emplace_back(time, c10::make_intrusive<Notification>());
-  c10::intrusive_ptr<Notification> notification =
-      fetch_notifications_.back().second;
+  c10::intrusive_ptr<Notification> notification;
+  {
+    std::unique_lock<std::mutex> lock_fetch(fetch_notifications_mutex_);
```
PaulZhang12, line 30: Why is this lock needed? Line 20 grabs the lock mu_, so this function should be thread safe?

contrib/dynamic_embedding/src/tde/ps.cpp
```diff
@@ -136,22 +142,30 @@ void PS::Evict(torch::Tensor ids_to_evict) {
...
   }
   notification.Wait();
 }
 
 void PS::SyncFetch(int64_t time) {
-  while (!fetch_notifications_.empty()) {
-    auto& [t, notification] = fetch_notifications_.front();
-    if (t != time && time >= 0) {
+  std::unique_lock<std::mutex> lock(
+      fetch_notifications_mutex_, std::defer_lock);
```
PaulZhang12, line 155: I assume SyncFetch and Fetch are not meant to be called concurrently in general? Let's use the mu_ lock used in Fetch as well here, which means no need to create a new lock. Unless we want more granularity on threading here, where we want the unlocking between waiting for each notification. Curious to your thoughts

contrib/dynamic_embedding/src/tde/ps.cpp
sherlockkenan: Although Fetch grabs the mu_ lock, the notification is accessed in SyncFetch without the lock (mu_), so the access to notification is not thread-safe. Here, a finer-grained lock is chosen to independently protect the notification instead of adding the mu_ lock during the SyncFetch process. The reason for this will be explained below.

contrib/dynamic_embedding/src/tde/ps.cpp
sherlockkenan: normally , SyncFetch and Fetch would be called concurrently, because Fetch returns an asynchronous handler and then continues to the next fetch. The asynchronous handler will be waited on by other threads calling SyncFetch. Therefore, using the mu_ lock would cause SyncFetch and Fetch to be completely mutually exclusive, which means that Fetch cannot be executed while waiting for SyncFetch, leading to a decline in performance. Therefore, a finer-grained lock is created here to protect only the access to the notification.

PaulZhang12: This makes sense, I can see why we want the granularity

contrib/dynamic_embedding/src/tde/ps.cpp
PaulZhang12: Do we even need mu_ lock then? Is that something we can get rid of? Could be a follow up pr

PaulZhang12 (dismissed):

contrib/dynamic_embedding/src/tde/ps.cpp
sherlockkenan: Based on my understanding, the primary purpose of mu_ lock is to ensure that the calls to Fetch and Evict are mutually exclusive, as variables such as cache_ids_to_fetch_or_evict_ and shards_ are shared between them. Although the current implementation in the Python side actually guarantees that they won't be called simultaneously, the way they are called from the Python side may change in the future. Therefore, the mu_ lock might still be necessary.

contrib/dynamic_embedding/src/tde/ps.cpp
sherlockkenan: thanks! By the way, I'm wondering if there's anything I can do since the process is stuck at this point.
image

contrib/dynamic_embedding/src/tde/ps.cpp
sherlockkenan: Additionally, we've also encountered performance issues .  The main problem is that both fetch and evict access data by reading or writing individual tensors one by one, which results in a large number of small GPU operators (on the same order of magnitude as the number of keys).  for example https://github.com/pytorch/torchrec/blob/main/contrib/dynamic_embedding/src/tde/ps.cpp#L58
By merging these into a larger batch operator, we can achieve a significant performance improvement. We'll clean up the code and submit it as a new mr later.

contrib/dynamic_embedding/src/tde/ps.cpp
PaulZhang12: Oops, approved sorry

contrib/dynamic_embedding/src/tde/ps.cpp
```diff
@@ -25,9 +25,12 @@ c10::intrusive_ptr<FetchHandle> PS::Fetch(
   if (cache_ids_to_fetch_or_evict_.empty()) {
     return c10::make_intrusive<FetchHandle>(time, c10::intrusive_ptr<PS>());
   }
-  fetch_notifications_.emplace_back(time, c10::make_intrusive<Notification>());
-  c10::intrusive_ptr<Notification> notification =
-      fetch_notifications_.back().second;
+  c10::intrusive_ptr<Notification> notification;
+  {
+    std::unique_lock<std::mutex> lock_fetch(fetch_notifications_mutex_);
+    fetch_notifications_.emplace_back(time, c10::make_intrusive<Notification>());
```
PaulZhang12, line 31: Formatting, please make:

<pre>
fetch_notifications_.emplace_back(
    time, c10::make_intrusive<Notification>());
notification = fetch_notifications_.back().second;
</pre>

PaulZhang12:
Please address CI failures and then you are good

sherlockkenan:
> Please address CI failures and then you are good

thanks! sorry for the failures and I've fixed them and pushed the changes to the merge request. However, it seems that the workflow requires approval again.

PaulZhang12:
CI looks good, previously flaky test. Feel free to merge!

PaulZhang12 (approved):

PR #1978, opened 2024-05-10
Integrate heterogenous auto planner to TGIF DI

gnahzg:
Summary:
As titled

Integrate TorchRec autoplanner for uneven sharding through option use_trec_hetero_auto_planner. The autoplanner would only be used when user explicitly set so.

Continue testing would be done before we fully migrate to use TorchRec Hetero Autoplanner. Aiming for timeline end of H1 2024

Differential Revision: D56203531


jingsh (approved):
lgtm

could you fix the pyre errors?

Issue #1614, opened 2024-01-09
Cannot compute the gradients of EmbeddingBagCollection.


ZhuYuJin:
I try to test the demo scripts with the following command.
```torchx run -s local_cwd dist.ddp -j 1x1 --script test_installation.py```

I try to print the gradients of embedding_bag_collection. I can observe the gradients of mlp linear layer. However, the gradients of embedding_bag_collection seem to be None.
![image](https://github.com/pytorch/torchrec/assets/5879410/5e3ae0ef-5320-4ebd-b997-9f88dce3b987)

**Comment by henrylhtsang** [comment]
We can't really look into the gradients. https://github.com/pytorch/torchrec/issues/1293

I guess Colin was saying if you switch to dense compute kernel, then you can see it. Though I haven't tried it myself.

**Comment by Ye-Tian-Zero** [comment]
@colin2328 Hi, as I also encountered the same problem, and after searching some issues, I found that the grad can be provided in dense compute kernel.
But I think dense compute kernel was diabled due to this commit:
https://github.com/pytorch/torchrec/commit/35c05fa00d2ca7e7b6fd577374560a5174be2675

Do you have any other ways to run a embedding collection in dense mode?

**Comment by Ye-Tian-Zero** [comment]
I found one way is to implement my own Sharder
```python
class EmbeddingCollectionSharderDense(EmbeddingCollectionSharder):
        def compute_kernels(self, sharding_type: str, compute_device_type: str) -> List[str]:
            return super().compute_kernels(sharding_type, compute_device_type) + [EmbeddingComputeKernel.DENSE.value]
```

henrylhtsang:
@Ye-Tian-Zero see to my answer in https://github.com/pytorch/torchrec/issues/1741

tldr: for non-data parallel sharding type, you get better performance by using FUSED. 

We didn't disable DENSE. We disabled DENSE in most tests, since its slower than FUSED.

For data parallel, you must use DENSE.

Ye-Tian-Zero:
@henrylhtsang Hi, thank you, but I think dense compute kernel was disabled here:

https://github.com/pytorch/torchrec/commit/35c05fa00d2ca7e7b6fd577374560a5174be2675#diff-72b4a4f205f4de558b2f5731a6697ae80483ee532eb2012d59999e5d7c462200L302-R302

which means Dense kernel was no longer a valid option for non-data-parallel module.

Ye-Tian-Zero:
[...]

PR #2012, opened 2024-05-17
add dense kernel for all sharding types

Ye-Tian-Zero:
Summary:
Allow to use dense kernel for all kinds of sharding types.

More details: https://github.com/pytorch/torchrec/issues/1614#issuecomment-1980378724


henrylhtsang:
@colin2328 what do you think? do you know why we disabled it in the first place?

Ye-Tian-Zero:
@colin2328  Any suggestion?

PR #2051, opened 2024-05-28
unify seq rw input_dist

YazhiGao:
Summary:
* unify unnecessary branching for input_dist module
* fx wrap some splits for honoring non-optional points.

Differential Revision: D57876357


jingsh (approved):
can you fix the pyre, it's an easy fix

PR #2115, opened 2024-06-14
use tcpstore based rdzv

kurman:
Summary:
Addressing test breaks introduced by D58180193
Instead of using filestore, using default TCPStore based store

Differential Revision: D58538848


kurman:
@henrylhtsang: a) are CI tests broken? b) Should I get an approval on PR as well before landing in diff tool?

henrylhtsang:
@kurman yes ci is broken. I think you can just land the diff, and the pr should auto close.

PR #2138, opened 2024-06-20
Give user option for multiplier to control preference for rank 0 or rank 1-N for Heteroplanner uneven sharding

gnahzg:
Summary: This diff add a multiplier to penalize non zero shards (parameter server shard) commum cost. This essentially give user a toggle to prefer rank 0 vs rank 1-N

Differential Revision: D58755047


jingsh (approved):
lgtm

PR #2217, opened 2024-07-09
Fix set_device propogation for MergePooledEmbedding Modules

PaulZhang12:
Summary: Fixing device propogation for setting EmbeddingsAllToOne device for TGIF verification

Differential Revision: D59544263


jingsh (approved):
lgtm

PR #2263, opened 2024-07-31
Cast tensor dtype to long in _unwrap_kjt_for_cpu

ljyuva83:
Summary: Follow up for: D60489592. The tensor dtype for indices could be casted to int32 by Preproc (e.g.: IGRankingTRPTrainPreproc) before lookup, this causes TBE lookup error in fbgemm kernel: https://fburl.com/everpaste/sqdb7rmt.

Differential Revision: D60535441


jingsh (approved):
duplicated changes?

also can you make sure the long type works with sharded ebc?

PR #2294, opened 2024-08-13
Sync .pyre_configuration.external files for those that got out of sync with last upgrade

connernilsen:
Summary:
This downgrades the .pyre_configurations in the following projects to keep their internal and external configurations synced.
- beanmachine/beanmachine/ppl
- tools/sapp
- torchrec
- torchx

The versions were chosen to match the most recent Pyre version in [pyre-check-nightly](https://pypi.org/project/pyre-check-nightly/#history).

Differential Revision: D61211101


yikaiMeta (changes requested):

PR #2426, opened 2024-09-25
Support global clipping for n-D model parallelism

yoyoyocmu:
Summary:
Add support for global clipping in the n-D parallelism case with following changes:
1. Add flag: `enable_global_grad_clip` to enable/disable global clipping.
2. Add the support to parse FSDP modules and find out the corresponding `process_group` or `DeviceMesh` that model parameter/gradient are sharded.
3. Accumulate the `norm` cross all model parallelism groups if global clipping is enabled.

Differential Revision: D62816452

### DIFF
--- a/torchrec/optim/clipping.py
+++ b/torchrec/optim/clipping.py
@@ -6,17 +6,18 @@
 # LICENSE file in the root directory of this source tree.
 
 # pyre-strict
-
+import logging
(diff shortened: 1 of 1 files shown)


y-sq (approved):

chuanhaozhuge (approved):
Reviewed at Meta internally.

PR #2438, opened 2024-09-27
Clean up legacy comments

sphinx07:
Summary: In D63148032， we make non-collections traceble, remove the legacy comments to avoid confusion.

Reviewed By: ZhengkaiZ

Differential Revision: D63552522


ZhengkaiZ (approved):
LGTM

PR #2511, opened 2024-10-23
Fix inference example and polish CMAKE paths and README

PaulZhang12:


agunapal (commented):
Is gRPC needed  for TorchRec or this for the server being shown in the example?


We do have a pytest for dlrm in TorchServe. Might be also worth showing a script example of torchrec inference working? I am working on a standalone script. Will share once ready

PR #2654, opened 2024-12-23
Update gcc version for FBGEMM install in CI

sarckk:
Summary:
TorchRec CI currently is failing with issues on incompatible GLIBCXX version. The cause is that FBGEMM now requires g++ 11.1+ for building binaries that reference GLIBCXX_3.4.29 (as of https://github.com/pytorch/pytorch/pull/141035)

As recommended in https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash and https://github.com/pytorch/FBGEMM/issues/3423, install GCC using conda to control glibcxx version being used.

Differential Revision: D67607624

### DIFF
--- a/.github/scripts/install_fbgemm.sh
+++ b/.github/scripts/install_fbgemm.sh
@@ -12,7 +12,7 @@ echo "CHANNEL"
 echo "$CHANNEL"
 
 if [ "$CHANNEL" = "nightly" ]; then
-    ${CONDA_RUN} pip install fbgemm-gpu --index-url https://download.pytorch.org/whl/nightly/"$CU_VERSION"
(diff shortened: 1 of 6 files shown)


huydhn:
I think I have a fix for this upstream in fbgemm (where the issue comes from) https://github.com/pytorch/FBGEMM/pull/3528, so let's hold on to see if we need this PR

huydhn:
@sarckk Sorry for all the delay.  I want to double check if this issue is still relevant.  One thing is that FBGEMM and TorchRec should not need to install a newer gcc version now because we have finished updating the build to use gcc-11.  I removed the step to install gcc in FBGEMM in https://github.com/pytorch/FBGEMM/pull/3528.

If there is still an issue somewhere, please ping me the link to that failure, I could take a look

PR #2671, opened 2025-01-08
[fbgemm_gpu] Set LD_LIBRARY_PATH in install script

q10:
- Set LD_LIBRARY_PATH in the fbgemm_gpu install script to fix the issue with fbgem_gpu being unable to locate libnvrtc.so on start


TroyGarden:
thanks for the prompt fix.

PR #2682, opened 2025-01-14
fix ZCH inference input dist error

emlin:
Summary:
Serving test failed when ZCH sharding is enabled because embedding lookup request Out of Bound.

After compared with training sharding input_dist, found that the keep_orig_idx flag is True when calling training block bucketize kernel, but not set in inference. So this diff is to make that consistent between training and inference, so the ID distribution works the same between training and inference.

Differential Revision: D68123290


kausv (approved):
LGTM

PR #2693, opened 2025-01-22
Teach is_signature_compatible() to dig into similar annotations

aorenste:
Summary: D68450007 updated some annotations in pytorch. This function wasn't correctly evaluating `typing.Dict[X, Y]` and `dict[X, Y]` as the equivalent.

Differential Revision: D68475380


izaitsevfb (approved):
forward fix for https://github.com/pytorch/pytorch/pull/145175

failures are on trunk

PR #2736, opened 2025-02-10
Refactor itep logger

peterfu0:
Summary:
put interface in the torchrec module and scuba impl in fbgemm, because it seems all GenericITEPModule needs a logger but only the references of GenericITEPModule in fbgemm module use scuba.


peterfu0:
needs an update

PR #2764, opened 2025-02-25
Use an env variable to roll out model sharding cache

hstonec:
Summary: We got positive result from AI Labs for the cache but would like to quantify the gains in produciton. As torchrec is an open source lib, we cannot use JK so need to rely on an environment variable to control it.

Reviewed By: andywag

Differential Revision: D70078729


andywag (approved):

andywag (changes requested):

PR #2794, opened 2025-03-11
support quantize fp32 embedding tables

tiankongdeguiji:


tiankongdeguiji:
hi, @YazhiGao @kausv @PaulZhang12 @henrylhtsang could you take a look?

PR #2829, opened 2025-03-17
fix divide NoneType error when use mc-ebc and mean pooling

tiankongdeguiji:


tiankongdeguiji:
hi, @sarckk @iamzainhuda @henrylhtsang @PaulZhang12 could you take a look?

iamzainhuda:
ack, taking a look!

tiankongdeguiji:
hi, @iamzainhuda Are there any existing issues I should address?

iamzainhuda:
i am thinking if we need to create the input dist here @tiankongdeguiji, we just need to include the mean pooling creation and callback w.r.t the feature permute right?

tiankongdeguiji:
> i am thinking if we need to create the input dist here @tiankongdeguiji, we just need to include the mean pooling creation and callback w.r.t the feature permute right?

yes

PR #2882, opened 2025-04-11
get rid of unncessary fx wrap in regrouping

ZhengkaiZ:
Summary: [torchrec] get rid of unncessary fx wrap in regrouping

Differential Revision: D72884977

### DIFF
--- a/torchrec/modules/regroup.py
+++ b/torchrec/modules/regroup.py
@@ -35,19 +35,6 @@ def _permuted_values(
 
 
 @torch.fx.wrap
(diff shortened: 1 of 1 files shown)


cp2923 (approved):
Double check, is there a test plan? Or existing UTs are sufficient

ZhengkaiZ:
> Double check, is there a test plan? Or existing UTs are sufficient

Yes, existing test case already covered it

jingsh (approved):

PR #2895, opened 2025-04-18
Fix GAUC not calculated with weights

xinyuanzzz:
Summary: The gAUC score is lower than expected e.g. https://fburl.com/mlhub/vljz497c. In ig, if a label presence is false, the corresponding weight is set to 0. It should not be considered when calculating gAUC.

Differential Revision: D73231152


xinyuanzzz:
The failed tests are because of [D72010614](https://www.internalfb.com/diff/D72010614). Will rebase after the fix lands.

PR #2968, opened 2025-05-13
Support DI sharding for FPE_EBC

hannaxu:
Summary:
Support models that have FeatureProcessedEmbeddingBagCollection.

These changes make sure we add QuantFeatureProcessedEmbeddingBagCollectionSharder as a recognized sharder, handle multiple envs needed for specifying DI sharding, and propagate TBE properly when processing the sharding plan.

Differential Revision: D74671655

### DIFF
--- a/torchrec/distributed/quant_embeddingbag.py
+++ b/torchrec/distributed/quant_embeddingbag.py
@@ -430,7 +430,7 @@ class ShardedQuantFeatureProcessedEmbeddingBagCollection(
         self,
         module: EmbeddingBagCollectionInterface,
         table_name_to_parameter_sharding: Dict[str, ParameterSharding],
-        env: ShardingEnv,
+        env: Union[ShardingEnv, Dict[str, ShardingEnv]],  # support for hybrid sharding
         fused_params: Optional[Dict[str, Any]] = None,
         device: Optional[torch.device] = None,
         feature_processor: Optional[FeatureProcessorsCollection] = None,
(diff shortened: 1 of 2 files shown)


torchrec/distributed/quant_embeddingbag.py
```diff
@@ -462,11 +462,30 @@ def __init__(
...
+                # throws if not all shards only have the expected device type
+                shard_device_type = get_device_from_sharding_infos(embedding_configs)
+                if isinstance(shard_device_type, tuple):
+                    raise RuntimeError(
+                        f"Sharding across multiple device types for FeatureProcessedEmbeddingBagCollection is not supported yet, got {shard_device_type}",
+                    )
+                assert (
+                    shard_device_type == expected_device_type
+                ), f"Expected {expected_device_type} but got {shard_device_type} for FeatureProcessedEmbeddingBagCollection sharding device type"
+
```
jingsh, line 479: since we assert here and the `expected_device_type` is always `cuda`, why don't we keep the signature of the method's env as a single `ShardingEnv`, and just pass the `cuda`'s sharding evn to this?

jingsh (changes requested):

torchrec/distributed/quant_embeddingbag.py
hannaxu: While right now FPE is used on cuda, we can support it on cuda/cpu in the future. Performing the check here allows us to de-duplicate the logic so we can ensure there's one sharding type for FPE for all sharding paths.

PR #3122, opened 2025-06-20
Add round-trip test for KeyedJaggedTensor to_dict and from_dict

vaishnavirao24:
This PR adds a focused unit test to verify that converting a KeyedJaggedTensor to a dictionary and reconstructing it using .from_dict() behaves as expected.

Specifically, the test:

Creates a KeyedJaggedTensor

Converts it to a dictionary using .to_dict()

Rebuilds it via .from_dict()

Asserts that all key components — values, weights, keys, and offsets — are correctly preserved

✅ This test ensures that future changes won't silently break serialization, deserialization, or downstream logic relying on these methods.

 No production code is modified.

Let me know if you'd like any revisions. Thanks for reviewing!


torchrec/sparse/tests/test_keyed_jagged_tensor.py
```diff
@@ -53,6 +53,29 @@ def test_key_lookup(self) -> None:
...
+            offsets=offsets,
+        )
+
+    kjt_dict = kjt.to_dict()
+    kjt_round_trip = KeyedJaggedTensor.from_dict(kjt_dict)
+
+    self.assertTrue(torch.equal(kjt.values(), kjt_round_trip.values()))
+    self.assertTrue(torch.equal(kjt.weights(), kjt_round_trip.weights()))
+    self.assertEqual(kjt.keys(), kjt_round_trip.keys())
+    self.assertTrue(torch.equal(kjt.offsets(), kjt_round_trip.offsets()))
```
TroyGarden: indent error?

vaishnavirao24: Thanks for catching that — I’ll double-check the indentation, reformat the block with 4 spaces consistently, and re-push. I’ll also verify the test locally with pytest before re-running CI.

TroyGarden:
the changes and revisions make no sense to me, suspecting an ai bot.

PR #3370, opened 2025-09-11
multiplethread pipeline

zhaozheng09:
we have a host bound model on A100
I found that there would be a long wait when doing embedding key preprocessing, which would cause the subsequent CPU scheduling to not keep up with the GPU operation, resulting in insufficient GPU utilization.
so we move ```unique and wait op``` from main thread to other thread .
before:
image
after:
image

I have roughly proposed a multi-threaded solution, but I don't know if it is feasible. If it is feasible, I will provide additional and complete code.

### DIFF
--- a/torchrec/distributed/train_pipeline/train_pipelines.py
+++ b/torchrec/distributed/train_pipeline/train_pipelines.py
@@ -60,6 +60,7 @@ from torchrec.distributed.train_pipeline.utils import (
     _prefetch_embeddings,
     _rewrite_model,
     _start_data_dist,
+    _prepare_data_dist,
     _start_embedding_lookup,
     _to_device,
     _wait_for_batch,
@@ -646,6 +647,10 @@ class TrainPipelineSparseDist(TrainPipeline[In, Out]):
             logger.info("fill_pipeline: failed to load batch i+1")
             return
 
+    def _data_processing_worker(self) -> None:
+        if len(self.batches) >= 2:
+            self.start_sparse_data_dist(self.batches[1], self.contexts[1], async_op=True)
+
     def _wait_for_batch(self) -> None:
         batch_id = self.contexts[0].index if len(self.contexts) > 0 else "?"
         with record_function(f"## wait_for_batch {batch_id} ##"):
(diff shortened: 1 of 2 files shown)


TroyGarden:
Hi @zhaozheng09 , thanks for the PR. wondering if you can share the trace files (before vs after) so that we can have better context. Usually the data preprocessing (PreProc) is done in a remote worker, but we'll consider your use case as well.

zhaozheng09:
> Hi @zhaozheng09 , thanks for the PR. wondering if you can share the trace files (before vs after) so that we can have better context. Usually the data preprocessing (PreProc) is done in a remote worker, but we'll consider your use case as well.

<File size too big: 25 MB are allowed, 1066 MB were attempted to upload. > Are there other ways to discuss this? or do I need to add any additional information?

TroyGarden:
usually you can gzip the file, and if it's still too large maybe you can try send me via email (hhy@meta.com). please let me know if the email still doesn't work. my main concern is that the PreProc logic is usually handled outside of torchrec, but your diff introduces extra dependency on dataloader (worker).

zhaozheng09:
> usually you can gzip the file, and if it's still too large maybe you can try send me via email ([hhy@meta.com](mailto:hhy@meta.com)). please let me know if the email still doesn't work. my main concern is that the PreProc logic is usually handled outside of torchrec, but your diff introduces extra dependency on dataloader (worker).

I'm trying to create a smaller timeline.

However, I don't quite understand the meaning of this statement:
"My main concern is that the PreProc logic is usually handled outside of TorchRec, but your diff introduces extra dependency on dataloader (worker)."

I believe my intention is quite clear:
The main thread needs to synchronously wait for the unique D2H (Device-to-Host) transfer to complete.

My optimization:
By moving this host-bound operation (waiting for D2H) to a secondary thread, we enable the host to:

Proactively schedule backward kernels to the GPU earlier
Achieve parallelism between:
Host-side operations (D2H waits)
GPU-side operations (backward kernel execution)

zhaozheng09:
[before.tar.gz](https://github.com/user-attachments/files/22660947/before.tar.gz)
[after.tar.gz](https://github.com/user-attachments/files/22660949/after.tar.gz)
@TroyGarden

TroyGarden:
sorry for the very late response. your multi-threading change on the sparse-data-dist pipeline has potential side effect:
1. you are moving the input-split dist to a separate thread, which involves the nccl all_to_all sync collective call
2. there are other all_to_all calls on the main thread. 
3. the risk is that in multithreading environment, the execution order isn't guaranteed on each rank, and it potentially causes trainers hanging or failing.

My suspicion is that this change might just work for your model (might have very clear runtime signature so the order is always certain, or with some barrier/sync to ensure the order). However, sparse-data-dist is a widely used pipeline we won't be able to educate all the user to follow the same practice. 

My suggestion is to create/subclass a new pipeline class in the train_pipeline/experimental_pipelines.py with your change.

TroyGarden:
I'm going to close this PR, feel free to re-open it if you have other thoughts.

PR #3498, opened 2025-10-31
Enable fused path on Segment NE

iamzainhuda:
Summary:
# Motivation
T242704386 mentioned that Segment NE is not compatible of FUSED metrics compute, which can bring efficiency win from [post](https://fb.workplace.com/groups/429376538334034/permalink/1474708170467527)

# Solution
Run group by group NE computation on tensors of all tasks.

# Compatibility
#thanks to the suggestion from ge0405, we use backward compatible passing of compute mode using bool to avoid cyclic dependencies.

Differential Revision: D85879827


axeisghost:
Captured in #3499

PR #3562, opened 2025-11-19
Optimize train_pipeline: decouple pipeline processing and add overlap thread for CPU bound scenarios

leeq2016:
This PR decouples the pipeline-related processing from the main training workflow in the train_pipelinemodule. 

It also provides an option that, when enabled, overlaps the pipeline processing with the main training process to accelerate host-side performance in case of CPU bottlenecks.

'P' as pipeline process
'T' as train workflow
Before:
training: |--P--| |--T--| |---P---| |-----------T(fwd)----------| |----P----| |-------------------------T(bwd)---------------------------|

After:
training: |--T--| |--------------T(fwd)----------------| |-------------------------T(bwd)---------------------------|
helper:   |--P--| |---P---| |----P----|

In CPU-bound scenarios, this can significantly reduce the overall time of a single iteration.

Regards :)

### DIFF
--- a/torchrec/distributed/train_pipeline/train_pipelines.py
+++ b/torchrec/distributed/train_pipeline/train_pipelines.py
@@ -10,6 +10,7 @@
 import abc
 import contextlib
 import logging
+import threading
 from collections import deque
 from dataclasses import dataclass
 from typing import (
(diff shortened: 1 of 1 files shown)


leeq2016:
before:
image

after:
image

These are the test results of a TorchRec development version I used. With my relatively poor CPU performance, I improved training performance by 30%+ by decoupling and parallel overlapping (the baseline had a significant bottleneck on the CPU).

PR #3891, opened 2026-03-18
Filter empty grads before clip_grad_norm_ with infinity norm

sourabhrohilla:
Summary:
Fix the second code path for the _foreach_norm infinity norm crash on empty
tensors (T260346508, T256567466). D94430621 fixed the TorchRec custom
`_batch_cal_norm` path, but when `len(self._sharded_params) == 0`,
`GradientClippingOptimizer.step()` delegates to PyTorch core's
`torch.nn.utils.clip_grad_norm_()` which does not filter empty gradient
tensors. When `norm_type=inf`, `_get_total_norm` passes empty tensors to
`torch._foreach_norm` which raises RuntimeError because max() over an empty
set has no identity element.

This diff filters out parameters with no gradients or empty gradients (numel==0)
before calling `clip_grad_norm_`, matching the same filtering done in `_get_grads()`.
Also adds regression tests for this specific code path.

Root cause: Interaction of D79128843 (TorchRec clipping refactor) and D91979475
(PyTorch _foreach_norm validation). The refactor removed the empty-list guard
and the PyTorch change made empty lists with inf norm an error.

Reviewed By: wz337

Differential Revision: D97129718

### DIFF
--- a/torchrec/optim/clipping.py
+++ b/torchrec/optim/clipping.py
@@ -120,11 +120,20 @@ class GradientClippingOptimizer(OptimizerWrapper):
                     p._local_tensor if isinstance(p, DTensor) else p
                     for p in self._replicate_params
                 ]
-                torch.nn.utils.clip_grad_norm_(
-                    parameters=replicate_params,
-                    max_norm=self._max_gradient,
-                    norm_type=self._norm_type,
-                )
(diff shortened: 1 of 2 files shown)


wz337 (approved):

PR #4326, opened 2026-06-06
Pass D=/I= as keyword args to faiss Index.search in two_tower example (#4326)

BernardZhao:
Summary:

The faiss Index.search torch overload declares its D/I output tensors as
keyword-only parameters; the retrieval two_tower example passed them
positionally, which the type checker rejects. Pass them as D=/I= keywords
(matches the faiss torch_utils runtime signature).

Reviewed By: angelayi, aeolyus, grievejia

Differential Revision: D107715937


grievejia (approved):

PR #4396, opened 2026-06-30
Allocate pairwise-distinct train pipeline streams

jeffdaily:
TorchRec's train pipelines allocate several CUDA streams (memcpy, data dist, prefetch, default) to overlap input distribution and prefetch with compute. Each was created independently with torch.cuda.Stream(), which on some backends can hand back streams that share underlying resources; work that was meant to overlap then serializes. This is most visible on backends with a small stream pool.

Route the pipeline's streams through a helper, _get_distinct_streams, that reserves each stream from the pool (torch.cuda.Stream(reserve=True)) so the set is pairwise distinct. It is backward-compatible: on a PyTorch without reservation support the reserve kwarg raises TypeError and the helper falls back to plain Stream() allocation, so older torch keeps working unchanged. A genuine pool-exhaustion error is not swallowed, so it still surfaces.

For full effect this depends on the PyTorch-side reservation support in pytorch/pytorch#188316 (torch.cuda.Stream(reserve=True) and the underlying pool reservation). Without that PR the helper falls back to non-reserved allocation, which on a small pool can still share resources; with it, each pipeline stream gets its own dedicated pool stream.

Test Plan:

The reservation mechanism and the TypeError fallback were validated on-device against a PyTorch build that has pytorch/pytorch#188316:

```
python - <<'EOF'
import torch
m = torch.get_device_module(torch.device("cuda", 0))
torch.cuda.set_device(0)
s = [m.Stream(priority=-1, reserve=True) for _ in range(2)]
assert s[0].cuda_stream != s[1].cuda_stream            # reserved -> distinct
try:
    m.Stream(priority=0, not_a_real_kwarg=True)         # unsupported kwarg
    raise SystemExit("expected TypeError")
except TypeError:
    pass                                                # == old-torch fallback trigger
assert [m.Stream(priority=0) for _ in range(2)]         # plain fallback still works
EOF
```

The existing train-pipeline tests construct these pipelines and exercise the stream-allocation path in CI:

```
python -m pytest torchrec/distributed/train_pipeline/tests/test_train_pipelines.py
```

Authored with assistance from Claude (Anthropic).

### DIFF
--- a/torchrec/distributed/train_pipeline/train_pipelines.py
+++ b/torchrec/distributed/train_pipeline/train_pipelines.py
@@ -155,6 +155,32 @@ except Exception:
 
 logger: logging.Logger = logging.getLogger(__name__)
 
+
+def _get_distinct_streams(
+    device: torch.device, num_streams: int, priority: int = 0
+) -> List[Optional[torch.Stream]]:
+    """Allocate ``num_streams`` device streams guaranteed to be pairwise distinct.
+
+    Requesting streams independently can leave them sharing one underlying stream
+    on some backends, which serializes work that was meant to overlap. Reserving
+    each stream from the pool keeps a pipeline's streams separate. Falls back to
+    plain ``Stream`` allocation on devices or torch builds without reservation
+    support (e.g. MTIA, or a torch without ``Stream(reserve=...)``).
+    """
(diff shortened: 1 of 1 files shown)


TroyGarden:
no use case provided.

PR #4428, opened 2026-07-15
Minimal repro: multi-host all_to_all lowering failure (world=32)

kausv:
Summary:
Minimal, standalone reproduction of a torch_tpu multi-host all_to_all lowering failure. A plain `torch.distributed.all_to_all_single` on the `tpu` device (`repro_collective_lowering.py`) fails to lower at world=32 (multi-host) and aborts every rank; the identical op at world=8 (single host) lowers cleanly, so the failure is specific to the world=32 cross-host device mesh.

Latest verified run (world=32):

```
repro: world_size=32
loc("distributed.all_to_all_single/distributed.all_to_all_single"): error: replica id #0 seen more than once
LLVM ERROR: Failed to infer result type(s):
    @  ...  mlir::detail::reportFatalInferReturnTypesError()
=> Signal 6 (SIGABRT) received on all 32 ranks (torch.distributed.elastic ChildFailedError)
```

Differential Revision: D111787064

### DIFF
--- a/torchrec/experimental/torch_tpu/repro_collective_lowering.py
+++ b/torchrec/experimental/torch_tpu/repro_collective_lowering.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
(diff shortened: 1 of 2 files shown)


wconstab:
This might be unrelated, but i've seen "error: replica id #{id} seen more than once" (id can be different values) caused by some kind of networking stack error that only seems to get fixed when we reboot the machines.

PR #4477, opened 2026-07-30
Pass the training device to get_device_module() in TorchRec train pipeline (#4477)

georgehong:
Summary:

`torch.get_device_module()` called with no argument resolves the default device module, and that default is moving to build-time detection (it returns the compiled-in accelerator even when no accelerator device is available at runtime), please see: https://github.com/pytorch/pytorch/pull/190751. Pass the pipeline's `self._device` so the `synchronize()` call targets the module for the actual training device and does not trigger accelerator initialization on hosts without a usable device.

Differential Revision: D114277607

### DIFF
--- a/torchrec/distributed/train_pipeline/train_pipelines.py
+++ b/torchrec/distributed/train_pipeline/train_pipelines.py
@@ -1843,7 +1843,9 @@ class TrainPipelineSemiSync(TrainPipelineSparseDist[In, Out]):
             del context
 
         if len(self.batches) >= 1 and not is_semi_sync:
-            torch.get_device_module().synchronize()  # needed to avoid race condition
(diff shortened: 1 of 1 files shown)


georgehong:
It looks like the single failing signal is unrelated (and seems to be happening on trunk): 
> The single CI failure is unrelated to this change. The failing shard (py314t free-threaded CUDA) dies during env setup at the `import fbgemm_gpu` check with an ABI mismatch (`undefined symbol: torch::autograd::_wrap_outputs`) — a fbgemm_gpu/torch nightly incompatibility that happens before any test runs, so the changed code is never exercised. All other shards, including the non-free-threaded 3.14 CUDA job, are green.

PR #4494, opened 2026-08-05
Revert D113962249: Gate optimizer scratch-buffer freeing behind a flag

doIIarplus:
Summary:
D113962249 added `OptimizerMemoryStashingConfig.free_scratch_buffers` to gate
freeing/re-allocating DistributedShampoo's `_global_dist_buffer` (the all-gather
scratch buffer for per-rank search directions) during the optimizer-state stash
window.

That gate only needed to exist because the buffer's lifecycle was routed through
torchrec's `MemoryStashingManager` and the APS MB train pipeline.
`_global_dist_buffer` is pure scratch that DistributedShampoo allocates, fills,
and consumes entirely within `update_params()` -- it never leaves the optimizer
and is not part of `state_dict`. Splitting its free/re-alloc across three
codebases forced a cross-repo `scratch_buffers()` protocol, optimizer-wrapper
fan-out, an extra restore-callback list in the stashing manager, a
`restore_scratch_buffer=False` special case on the checkpoint path, and this APF
config flag -- all to coordinate something that never crosses a module boundary.
It also coupled this feature's rollout to the persistent-state stashing rollout,
and only worked for pipelines that happen to use `MemoryStashingManager`.

Reverting the full stack (this diff, then D113962248, then D113962247). The
feature is being re-implemented entirely inside
`hpc/optimizers/distributed_shampoo/dev/`, driven by a new
`DDPDistributedConfig.free_dist_buffer_between_steps` flag, with the free /
re-alloc confined to the distributor that owns the buffer.

Bottom of a 3-diff revert stack.

Original commit changeset: 3d74a0bcb80d
Original Phabricator Diff: D113962249

Differential Revision: D114959515

### DIFF
--- a/torchrec/distributed/memory_stashing.py
+++ b/torchrec/distributed/memory_stashing.py
@@ -1079,7 +1079,6 @@ class MemoryStashingManager:
         optimizer: torch.optim.Optimizer,
         sync_event: Optional[torch.cuda.Event] = None,
         num_slices: int = 1,
-        free_scratch_buffers: bool = True,
     ) -> Tuple[
         Callable[[Optional[torch.Tensor]], None],
         Callable[[Optional[torch.Tensor]], None],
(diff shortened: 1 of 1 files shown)


wz337 (approved):

PR #4642, opened 2026-08-31
Preserve index dedup in sequence outputs (#4642)

TroyGarden:
Summary:
Design review: [Packed JaggedTensor design](https://docs.google.com/document/d/1jTNkumyn767YKPgazwZ6ogkRCPQPjTeS83ibLgOm0Oo/edit?tab=t.0#heading=h.cf4pvk51zsam)

## 1. Context

Sharded EmbeddingCollection can deduplicate input indices before the embedding lookup, but construct_jagged_tensors currently expands the deduplicated rows with reverse_indices while building per-feature outputs. This restores the original logical shape immediately, so downstream consumers cannot retain the reduced embedding storage or defer re-duplication.

## 2. Design

1. Immutable representations: A JaggedTensor is either materialized or packed for its entire lifetime. Calling values() on a packed JT and packed_values() on a materialized JT raises, preventing accessors from silently changing object state.
2. Explicit materialization boundary: materialize() returns a separate, normal JaggedTensor by applying reverse_indices. The packed source remains unchanged, so logging, pytree flattening, and record_stream behavior do not depend on accessor order.
3. Sharder-owned opt-in: Add use_packed_jagged_tensor to EmbeddingCollectionSharder and thread it through ShardedEmbeddingCollection. Require index dedup when enabled and leave the local EmbeddingCollection API unchanged.
4. Explicit construction paths: Keep construct_jagged_tensors as the existing materialized path and add construct_packed_jagged_tensors for packed outputs. ShardedEmbeddingCollection selects the appropriate function at the call site.
5. Representation invariants: Reject packed construction without reverse indices and reject column-wise sharding, where duplicate embedding names require concatenating column shards rather than sharing a single packed row address space.

## 3. Results

1. Index-deduplicated embeddings remain packed after Sharded EmbeddingCollection output distribution until a consumer explicitly calls materialize().
2. Materialization preserves gradient accumulation through index_select without mutating the packed source.
3. Packed state remains stable across logging, device transfer, copying, stream recording, size accounting, and pytree round trips.
4. VBE packed outputs materialize to the same values, lengths, and weights as the existing eager expansion path.

## 4. Compatibility and scope

1. The sharder option defaults to False, so existing distributed and local EmbeddingCollection behavior is unchanged.
2. Packed output is an explicit contract: packed-aware consumers use packed_values() and reverse_indices(); consumers needing a regular JT call materialize().
3. Pytree unflattening accepts older four-field representations. Materialized JTs keep the legacy schema, while packed JTs carry reverse indices as a fifth field.
4. Packed output is intentionally limited to index-deduplicated, non-column-wise ShardedEmbeddingCollection configurations.

## 5. Changes

1. torchrec/distributed/embedding.py: Add and validate the sharder-level packed-output option, propagate it through the output awaitable, and select the packed or materialized constructor at the call site.
2. torchrec/modules/utils.py: Add construct_packed_jagged_tensors, including VBE reverse-index reordering, while retaining construct_jagged_tensors as the existing materialized path.
3. torchrec/sparse/jagged_tensor.py: Add strict packed accessors, explicit non-mutating materialization, packed-aware logging, and lifecycle support for reverse indices.
4. Tests: Cover packed address-space sharing, explicit materialization and gradients, stable logging and pytree behavior, VBE parity, invalid configurations, and distributed sharding integration.

# The two jobs below have trace:
* baseline: fire-gjin-train-baseline (trace: https://fburl.com/kernelhub/pmrwl0h5) 
* proposal: fire-gjin-train-uih-dedup (trace: https://fburl.com/kernelhub/ihliecwf)

Reviewed By: gjin7

Differential Revision: D117628626


sylvesterkaczmarek (commented):
Packed JTs still expose the normal `to_dense()` and `to_padded_dense()` APIs, but both call `self.values()`, which this PR now deliberately raises on packed storage. So enabling `use_packed_jagged_tensor` breaks callers that use logical conversion APIs without touching `values()` directly. Should those methods materialize internally, or explicitly reject/document packed JTs?

sylvesterkaczmarek (commented):
Re-reviewed current `3b89a93f`. The packed-JT API concern I raised is resolved by making the representation explicit and immutable: packed `values()` now fails with a clear `materialize()` requirement, while `materialize()` returns a separate normal JaggedTensor with logical values and preserved gradients. The distributed sequence-model call sites that need dense logical values now materialize explicitly, and the new packed/materialized regressions cover the boundary. No remaining blocker from my earlier review.

sylvesterkaczmarek (approved):
Re-reviewed the current exported head. The packed JaggedTensor contract remains explicit and immutable, materialization is still the boundary for logical values, and the distributed sequence-model paths materialize before dense conversion. My earlier concern remains resolved.

PR #4703, opened 2026-09-10
[perf] TrainPipelineSparseDist benchmark on 8x B200: 345.5 ms to 321.8 ms per step, mostly from an unread per-sample-weight gradient in the TBE backward

TarzanZhao:
## Summary

On 8 x B200, a step of `benchmark_train_pipeline` on the shipped `sparse_data_dist_base.yml` spends 18 ms computing a per-sample-weight gradient that autograd then discards, and another 13 ms moving the pooled-embedding gradient: three passes over about 6 GB per rank in the all-to-all backward, plus a staging copy in its forward. This PR changes the fused embedding-bag wrapper and the pooled all-to-all autograd functions; fbgemm, the planner, the pipelines and the benchmark are untouched, and neither commit adds a flag, an environment variable or a dependency. On the 8 B200s I measured, the step goes from 345.5 ms to 321.8 ms, **1.07x**, against a run-to-run spread of 0.5 % on that node. I have not measured other GPUs, other pipelines or other sharding types.

1. **Zero `feature_requires_grad` mask for the fused TBE** (`27815262`). fbgemm's backward computes `grad_indice_weights` whenever `indice_weights` is defined, and torchrec never sent the mask, so every step of a weighted lookup ran an 18 ms kernel whose result autograd discards. `BaseBatchedEmbeddingBag.forward` now passes an int32 zero mask, cached per device, when the KJT's per-sample weights do not require grad and the module is a `SplitTableBatchedEmbeddingBagsCodegen`, and the kernel returns early. Weighted lookups whose weights do require grad, such as those behind a feature processor, take the old path, as do the dense and SSD TBE variants. There is no opt-out, because the mask only tells fbgemm what autograd already knows.
2. **One-pass gradient recat and no staging copy for the default all-to-all** (`27691c7b`). `_recat_pooled_embedding_grad_out` (an upstream TODO, T87591139) turned the `[B, D_global]` gradient into the per-rank flat layout with `contiguous()`, then a `contiguous()` per rank, then `torch.cat`, which is three passes over about 6 GB per rank per step, 48 copy kernels and 3 cats, 11 ms; and `All2All_Pooled_Req.forward` copied its input into a second buffer that `DefaultAll2AllSingle` allocates with a plain `torch.empty`, 2 ms and 6 GB of live memory. Each column block is now copied once into its slot of a preallocated buffer with `torch._foreach_copy_`, the input may stay a non-contiguous view, and the staging copy is skipped for `DefaultAll2AllSingle` — so the in-flight all-to-all reads the caller's embedding tensor rather than a private copy. Comms that need a registered buffer keep the staging copy; neither part has an opt-out.

### Speedup Result

Measured on 8 x B200 in one node.

| | baseline (`45f3f4b3`) | this PR | change |
|---|---|---|---|
| training step, `gpu_runtime_p50_ms` / 50 on rank 0, two runs | 345.5 ms (344.9, 346.1) | 321.8 ms (322.6, 321.1) | **1.07x**, -6.9 % |
| QPS per rank | 94.9 k | 101.8 k | +7.3 % |
| peak allocated memory per rank | 33.9 GB | 31.1 GB | -2.8 GB |
| peak reserved memory per rank | 61.3 GB | 56.0 to 58.6 GB | |
| nvidia-smi peak, GPU 0 | 70.9 GiB | 65.8 GiB | |

Per-kernel, from traces taken in an earlier round against a different baseline, so these do not add up to the table above: `grad_indice_weights` 18.0 ms per step to below the top-12 kernel cutoff; the recat's 48 copies and 3 cats, 11.0 ms, to 76 copies, 5.6 ms; the all-to-all forward staging copy, 2.0 ms, gone.

Run-to-run spread was 1.2 ms (0.3 %) on the baseline and 1.5 ms (0.5 %) on this PR; within a run the three timed repetitions agree to 0.1 %. The two commits were measured together against that baseline. The split below is from that earlier round, one seeded run per commit against the untuned main on a different node, so its numbers sit on a different baseline and do not line up with the table above.

| commit | change | measured effect |
|---|---|---|
| | untuned main `45f3f4b3`, seeded, other node | 343.7 ms |
| `27815262` | zero `feature_requires_grad` mask | 327.4 ms, -16.3 ms |
| `27691c7b` | one-pass recat, no staging copy | 319.3 ms, a further -8.1 ms |

## Correctness Verification

I ran the baseline and this branch on the same 50 synthetic batches, 250 steps per rank, and compared 794 recorded values per rank on ranks 0 to 2 and 792 on ranks 3 to 7, 6342 per run: the loss and the prediction magnitude at every step, the mean and the mean of |w| of every local parameter shard after the last step, the sharding type, kernel and ranks of all 172 tables, and the input shapes and index counts of step 1. The tolerances were set before the runs: exact on the plan and on the step-1 shapes and index counts, 0.1 % on the loss and the prediction, and 0.1 % (plus 1e-6 absolute on the mean) on the parameter shards.

Nothing exceeded its tolerance: all 6342 values match bit for bit on all 8 ranks. Three seeded runs of the unmodified code in the earlier round were bit-identical on every checkpoint, so the noise floor on these values is 0. Two further compares tie this round to the earlier one: the earlier base record against this baseline, and the earlier branch record against this branch, each 6342 of 6342 bit-identical on a different node. The seeded pair also gives a timing, 346.4 ms for the baseline and 321.9 ms for this branch, which agrees with the unseeded runs above.

The recording code — the checkpoint hooks, their call sites in `benchmark_train_pipeline.py`, and a seeding hook, since this entry point has no seed and this torch nightly seeds from entropy — sits on a local branch that was not pushed and is not in this PR; the two source trees used for the compare are each identical to their arm on the two files this PR touches.

Repository tests, run on this branch and on the unmodified base through `unittest` (pytest is not installed in this environment): `torchrec.distributed.tests.test_comm`, which covers the all-to-all autograd functions the second commit changes, 7 tests OK on both; `torchrec.distributed.train_pipeline.tests.test_train_pipelines -k SparseDist`, 19 tests, the same 15 passing on both and the same 3 failures and 1 error in `TrainPipelineSparseDistCompAutogradTest` (compiled-autograd capture counts on this torch nightly) on the base as well. No test changes state between the two arms. This PR adds no test of its own.

<details>
<summary>Details: hardware, model, full command, traces</summary>

**Hardware and environment.** One node, 8 x NVIDIA B200 (sm_100, 183 GB each, NVLink between every pair), 2 x Xeon 8562Y+, 2 TB host RAM, driver 580.126.20, CUDA 13.0. Python 3.11.16, torch 2.15.0.dev20260908+cu130 (nightly wheel, NCCL 2.30.7), fbgemm_gpu 2026.9.8+cu130 (nightly wheel), torchrec at `45f3f4b3` or at this branch, put first on `PYTHONPATH`. All 8 GPUs read 0 MiB before every timed run.

**Measurement.** Against the unmodified `45f3f4b3`, both arms with the same wheels, the same environment and the same command; nothing was installed, set or flagged on either arm beyond housekeeping (`PYTHONPATH` to the checkout under test, a node-local `TMPDIR`, cache directories). I swept 8 settings on the base first (below) and kept none, so the baseline the speedup is measured against is the benchmark's own command, unchanged. Two runs per arm, interleaved baseline, branch, baseline, branch, `--num_batches=50`, 5 repetitions of 50 steps per process. The metric is the benchmark's own `gpu_runtime_p50_ms`, the rank-0 CUDA-event time of one 50-step repetition with the first and last of the 5 dropped, divided by 50. The benchmark has no seed and this torch nightly seeds from entropy, so each run draws its own synthetic batches and init; in the earlier round one unseeded base run in five came out 7 % fast, while none of the 22 unseeded runs of this round strayed (345.9 to 351.5 ms at 20 batches; 344.9 to 346.1 for the baseline and 321.1 to 322.6 for this branch at 50). The measurement covers this configuration only: TrainPipelineSparseDist, table-wise sharding with a column-wise `large_table`, world size 8, the default all-to-all. Other pipelines, sharding types, comms, world sizes and the compiled-autograd path were not measured.

**Model and job.** TestSparseNN as built by the shipped `sparse_data_dist_base.yml`, unchanged. What the numbers above depend on: 172 tables, of which 80 are in a *weighted* EmbeddingBagCollection (100 000 x 256, fused fbgemm TBE with EXACT_SGD) whose per-sample weights do not require grad, which is what makes the discarded `grad_indice_weights` kernel run; `large_table` (1 000 000 x 2048) sharded column-wise across 16 shards, every other table table-wise; 32 768 samples per rank per step over 100 float and 172 sparse features, about 89 M unweighted and 77 M weighted indices, which makes the pooled-embedding gradient a 32 768 x 46 080 fp32 tensor, about 6 GB each direction per rank; fp32 throughout with TF32 off, as the benchmark's `MultiProcessContext` sets it. Host RSS is 136 GB per rank on both arms, from the 50 pinned batches.

**Data.** None staged. The batches are synthetic and generated in host memory at startup, about 2.5 minutes of a roughly 6-minute run on both arms.

**The measured job**, both arms. The baseline arm is a checkout of upstream `45f3f4b3`, the other a checkout of this branch. The checkout under test goes first on `PYTHONPATH`, and the command runs from an output directory outside the checkout, because the benchmark writes relative to the working directory:

```
export PYTHONPATH=<checkout>          # 45f3f4b3 for the baseline arm, this branch for the other
cd <out dir>
python -m torchrec.distributed.benchmark.benchmark_train_pipeline \
    --yaml_config=<checkout>/torchrec/distributed/benchmark/yaml/sparse_data_dist_base.yml \
    --world_size=8 --num_batches=50 --num_profiles=0 \
    --profile_dir=<out dir> --output_json=True
```

Flags are the YAML's own except `--world_size=8` (YAML: 2), `--num_batches=50` (YAML: 20) and `--num_profiles=0`. The benchmark spawns the 8 processes itself (`run_multi_process_func`, backend `cpu:gloo,cuda:nccl`), so there is no torchrun. Rank 0 prints the result dict. The correctness compare ran the same command with the seeding hook at seed 0.

**Settings tried on the base and rejected**, all inside the 0.5 % floor, with a second run for anything within 2x of it; none kept: `--expandable_segments=true` (+0.3 % on the step, reserved memory 60.7 to 41.4 GB, allocated unchanged), `NCCL_CUMEM_ENABLE=0` (-0.46 %), `NCCL_NVLS_ENABLE=0` (+0.1 %), `OMP_NUM_THREADS=16` (-0.55 %), `FBGEMM_TBE_USE_TUNED_SEGMENT_LENGTHS_CTA_B200=1` (-0.5 %; the backward runs in the CTA-per-row kernel because every row is looked up about 79 times per step, and the gate only moves the warp/CTA split), `FBGEMM_BOUNDS_CHECK_INDICES_V2=1` (+0.05 %), `TORCH_NCCL_AVOID_RECORD_STREAMS=1` (-0.2 %). Excluded because they change what is computed: TF32, bounds checks off, the benchmark's sync and batch knobs, and any sharding, planner or pipeline change.

**Traces** (torch.profiler, earlier round, 3 steady steps of a 10-step pass after the timed region, all 8 ranks; open in https://ui.perfetto.dev). The baseline trace is the unmodified `45f3f4b3`, which is also this round's baseline, since the sweep kept nothing:

- baseline: https://drive.google.com/open?id=10Z9ftAyN-krcc2cIC9xzc9WqwMgUd3Um
- optimized: https://drive.google.com/open?id=1But_8zMbx2j5XK1cqrz_DPelzp3rdWfn

</details>


TarzanZhao:
I have picked the essential part of this PR, the zero `feature_requires_grad` mask for the fused TBE, into a smaller PR of its own, rebuilt on today's main and measured on its own: #4749. Putting this one back to draft in the meantime.

PR #4749, opened 2026-09-17
[perf] TrainPipelineSparseDist benchmark on 8x B200: 345.3 ms to 328.4 ms per step, from an unread per-sample-weight gradient in the TBE backward

TarzanZhao:
## Summary

fbgemm's fused TBE backward computes `grad_indice_weights` whenever `indice_weights` is defined, and torchrec never sends it a `feature_requires_grad` mask. In `benchmark_train_pipeline` on the shipped `sparse_data_dist_base.yml`, the 80 weighted tables are fed per-sample weights that are a leaf requiring no gradient, so every step runs that kernel and autograd discards what it writes. This PR passes a zero mask in exactly that case, and the kernel returns early. One file, 25 added lines, no flag, no environment variable, no dependency; fbgemm, the planner, the pipelines and the benchmark are untouched. On the 8 B200s I measured, the step goes from 345.27 ms to 328.41 ms, **1.051x**: 16.86 ms off a 345 ms step, 4.9 %, against a run-to-run spread of 0.31 % on the base and 0.24 % on this branch. I have not measured other GPUs, other pipelines or other sharding types.

`BaseBatchedEmbeddingBag.forward` now passes an int32 zero mask, cached per device, when the KJT's per-sample weights do not require grad and the embedding module is a `SplitTableBatchedEmbeddingBagsCodegen`. Weighted lookups whose weights do require grad, such as those behind a feature processor, take the old path, as do the dense and SSD TBE variants. There is no opt-out, because the mask only tells fbgemm what autograd already knows. The win is specific to that one case, a fused TBE fed per-sample weights that need no gradient. A model with no weighted tables, or one whose weights carry a gradient, sees no change.

### Speedup Result

Measured on 8 x B200 in one node, against upstream main at `297e5685`.

| | base (`297e5685`) | this PR | change |
|---|---|---|---|
| training step, `gpu_runtime_p50_ms` / 50 on rank 0, four runs | 345.27 ms (345.59, 344.89, 344.76, 345.83) | 328.41 ms (328.79, 328.00, 328.34, 328.50) | **1.051x**, -16.86 ms, -4.9 % |
| QPS per rank | 94.9 k | 99.8 k | +5.2 % |
| peak allocated memory per rank | 33.9 GB | 33.9 GB | none |
| peak reserved memory per rank | 60.1 to 61.9 GB | 60.3 to 61.8 GB | none |
| nvidia-smi peak, GPU 0 | 71.4 to 73.2 GiB | 71.6 to 72.5 GiB | none |

Run-to-run spread was 1.07 ms (0.31 %) on the base and 0.79 ms (0.24 %) on this PR, and within a run the three timed repetitions agree to 0.1 %. The effect is about 16 times that spread, and it separates the two on every pairing: the slowest run of this PR beats the fastest run of the base, 328.79 against 344.76, still 1.049x, and the widest pairing is 1.054x. This removes a kernel, not an allocation, so memory does not move.

## Correctness Verification

I ran the base and this branch on the same 50 synthetic batches at seed 0, 250 steps per rank, and compared 6342 recorded values per run across the 8 ranks: the loss and the prediction magnitude at every forward, every local parameter shard after the last step, the sharding plan, and the input shapes and index counts. The tolerances were set before the runs, exact on the plan and the shapes and 0.1 % elsewhere.

**All 6342 match bit for bit on all 8 ranks.** That is the expected result: the skipped kernel writes a gradient for a leaf that needs none, so nothing reads it and every kernel downstream runs on the same inputs in the same order. A record of main at `45f3f4b3` against one of today's `297e5685` is also bit-identical, so the roughly 80 commits in between did not touch this computation either.

The recording hooks sit on a local branch and are not in this PR. `test_train_pipelines -k SparseDist` gives the same result on this branch and on the base: 11 pass, 4 skip, 3 failures and 1 error, all in `TrainPipelineSparseDistCompAutogradTest` and all failing on the base too. `black` and `usort` pass on the changed file.

| recorded | base vs this PR | tolerance |
|---|---|---|
| loss and prediction magnitude, every forward, 8 ranks | every value bit-identical | 0.1 % relative |
| every local parameter shard after the last step | every value bit-identical | 0.1 % relative, 1e-6 absolute on the mean |
| sharding type, kernel and ranks of all 172 tables | identical | exact |
| input shapes and index counts | identical | exact |

<details>
<summary>Details: hardware, model, full command, measurement</summary>

**Hardware and environment.** One node, 8 x NVIDIA B200 with NVLink between every pair, driver 580.126.20. Python 3.11.16, torch 2.15.0.dev20260908+cu130 (nightly wheel), fbgemm_gpu 2026.9.8+cu130 (nightly wheel), torchrec at `297e5685` or at this branch, put first on `PYTHONPATH`. All 8 GPUs read 0 MiB before every timed run, and the run script refuses to start otherwise.

**Measurement.** Both versions ran the same wheels, the same environment and the same command; nothing was installed, set or flagged on either side beyond housekeeping (`PYTHONPATH` to the checkout under test, a node-local `TMPDIR`, cache directories). An earlier sweep of allocator, NCCL, OpenMP and fbgemm settings on the base kept none of them, so the baseline here is the benchmark's own command, unchanged. Four runs of each: two unseeded, interleaved base, branch, base, branch, and two seeded at seed 0, which are the runs the correctness compare came from and which carry their own timing. Each run is `--num_batches=50`, 5 repetitions of 50 steps per process. The metric is the benchmark's own `gpu_runtime_p50_ms`, the rank-0 CUDA-event time of one 50-step repetition with the first and last of the 5 dropped, divided by 50. The entry point has no seed and this torch nightly seeds from entropy, so each unseeded run draws its own synthetic batches and init; none of the four strayed, and the seeded runs agree with the unseeded pair to a third of a millisecond. The measurement covers this configuration only: TrainPipelineSparseDist, table-wise sharding with a column-wise `large_table`, world size 8, the default all-to-all. Other pipelines, sharding types, comms, world sizes and the compiled-autograd path were not measured.

**Model and job.** TestSparseNN as built by the shipped `sparse_data_dist_base.yml`, unchanged. What the numbers depend on: 172 tables, of which 80 sit in a *weighted* EmbeddingBagCollection (100 000 x 256, fused fbgemm TBE with EXACT_SGD) whose per-sample weights do not require grad, which is what makes the discarded `grad_indice_weights` kernel run; `large_table` (1 000 000 x 2048) sharded column-wise, every other table table-wise; 32 768 samples per rank per step over 100 float and 172 sparse features; fp32 throughout with TF32 off, as the benchmark's `MultiProcessContext` sets it.

**Data.** None staged. The batches are synthetic and generated in host memory at startup.

**The measured job**, both versions. One is a checkout of upstream `297e5685`, the other a checkout of this branch. The checkout under test goes first on `PYTHONPATH`, and the command runs from an output directory outside the checkout, because the benchmark writes relative to the working directory:

```
export PYTHONPATH=<checkout>          # 297e5685 for the base, this branch for the other
cd <out dir>
python -m torchrec.distributed.benchmark.benchmark_train_pipeline \
    --yaml_config=<checkout>/torchrec/distributed/benchmark/yaml/sparse_data_dist_base.yml \
    --world_size=8 --num_batches=50 --num_profiles=0 \
    --profile_dir=<out dir> --output_json=True
```

Flags are the YAML's own except `--world_size=8` (YAML: 2), `--num_batches=50` (YAML: 20) and `--num_profiles=0`. The benchmark spawns the 8 processes itself (`run_multi_process_func`, backend `cpu:gloo,cuda:nccl`), so there is no torchrun. Rank 0 prints the result dict. The correctness compare ran the same command with the seeding hook at seed 0.

</details>


TroyGarden:
@TarzanZhao can you provide more context about your use case? are you using weighted EBC with fixed weights?

TarzanZhao:
> @TarzanZhao can you provide more context about your use case? are you using weighted EBC with fixed weights?

Hi, I'm not using the repo for any specific use case. I'm actually building an agent that optimizes speed of open-source codebases automatically and safely. My motivation is that agents can now write better CUDA kernels than most humans, and there's a lot of low-hanging performance issues across open-source projects. So that's what I'm trying to tackle.

TroyGarden:
thanks for spending agent hours on our repo. Given the production safety requirements and lack of bandwidth to review agent-driven optimizations, I am closing this PR to avoid clogging your queue.

PR #4761, opened 2026-09-20
Fix absent weights in scripted KJT permutations

eric-gecheng:
An unweighted `KeyedJaggedTensor` can fail on a second `permute()` when a serialized model runs in native/C++ TorchScript:

```text
RuntimeError: Expected a proper Tensor but got None (or an undefined Tensor in C++) for argument #0 'self'
```

Some FBGEMM CUDA permutation kernels return a default-constructed, undefined Tensor in the optional weights slot. The resulting KJT carries this value into the next permutation, which attempts to use it as a valid tensor. Python operator boundaries convert undefined tensors to `None`, so ordinary eager tests can hide the failure. Nonempty inputs reproduce it as well; empty inputs are not required.

This change explicitly restores `permuted_weights = None` when the input KJT has no weights, before constructing the returned KJT. Valid weights retain their normal permutation behavior. The guard is included in newly scripted/exported models.

The regression test serializes two chained permutations and reloads them in a fresh process that registers native operator libraries without importing `fbgemm_gpu` or `torchrec`. It checks both permutation outputs on CPU and CUDA, with and without weights, empty and nonempty values, fixed and variable stride, and cached or generated permutation indices. Fixed stride covers int32/int64 lengths; the scripted variable-stride cases use int64 lengths.

## Test plan

- Before the fix, the new regression reproduced the error above on the second CUDA permutation. After the fix, all 48 native TorchScript cases pass.
- `python -m unittest torchrec.sparse.tests.test_keyed_jagged_tensor -v`: **60 tests passed**, using this source checkout with PyTorch 2.13.0+cu130 and FBGEMM-GPU 1.8.0.
- Additional local validation with standalone LibTorch 2.11.0: **48 CPU/CUDA cases passed**, including checks that both returned weights are real `None` IValues for unweighted inputs.
- Pre-commit checks passed for both modified files, including the configured ufmt/Black/usort formatting checks.


TroyGarden:
can you tell more about the use case, why/what is the second permute?
```
An unweighted KeyedJaggedTensor can fail on a second permute()
```

eric-gecheng:
> can you tell more about the use case, why/what is the second permute?
> 
> ```
> An unweighted KeyedJaggedTensor can fail on a second permute()
> ```

The use case is GPU inference of a TorchEasyRec recommendation model, exported as TorchScript and loaded by a C++ serving process. Its embedding path combines managed-collision (ZCH) ID remapping with a quantized EmbeddingBagCollection. 

The two permutations occur within the same forward pass, because these components require different feature orders:

1. Managed-collision feature ordering. ManagedCollisionCollection calls _mcc_lazy_init_inplace(), which uses features.permute() to select the relevant features and group them by embedding table. In this model, six features belong to four tables; three features share the item-ID table. Their resulting order is:

[user_id, user_cids, item_id, q_clk_iid_20, query_tokens_id, combo_qi_brand]
After remapping the IDs, the collection constructs a new KJT and carries over weights=features.weights_or_none()

2. Quantized embedding lookup ordering. The quantized EmbeddingBagCollection groups tables by pooling mode for its batched embedding operators. This model mixes SUM and MEAN pooling, so its forward() calls _permute_kjt() to obtain:

SUM:  [user_id, query_tokens_id, combo_qi_brand]
MEAN: [user_cids, item_id, q_clk_iid_20]

Relative to the managed-collision output, this requires permutation [0, 4, 5, 1, 2, 3], followed by split([3, 3]). This _permute_kjt() call is the “second permute” mentioned in the PR.

The input KJT is unweighted (weights=None). On the affected FBGEMM CUDA implementation, the first permutation returns an undefined Tensor in the optional weights slot. The managed-collision reconstruction preserves that value, so the second permutation receives it as though weights were present and fails inside fbgemm::permute_2D_sparse_data

I have also opened [pytorch/FBGEMM#6341](https://github.com/pytorch/FBGEMM/pull/6341) to fix the underlying native return value.

torchrec/sparse/jagged_tensor.py
```diff
@@ -3083,6 +3083,10 @@ def permute(
                 permuted_length_per_key_sum,
             )
 
+        # Native FBGEMM can return an undefined tensor for absent weights.
+        if self.weights_or_none() is None:
+            permuted_weights = None
```
TroyGarden, line 3090: what's the previous value for `permuted_weights` in your use case? probably want to mention it in your PR's summary, or just add a comment in the code.

TroyGarden (changes requested):
the test has low quality, please rewrite the test to match existing test style.

torchrec/sparse/jagged_tensor.py
eric-gecheng: permuted_weights was an undefined Tensor returned by FBGEMM for an unweighted input. Setting it to None prevents the next permute() from treating it as valid weights.
I added some comments in the code.

